PR #11, opened 2018-05-16
Create doc_requirements.txt

dwighthubbard:
This adds a doc_requirements.txt to tell readthedocs the packages it needs to install before generating the documentation.


doc_requirements.txt
```diff
@@ -0,0 +1,12 @@
+six
+pyparsing
+python-dateutil
+packaging
+requests
+xmlrpc2
```
cooperlees: Hopefully I can remove this and requests soon 💃

PR #16, opened 2018-05-16
Implement pep440 blacklisting

dwighthubbard:
This pull request has several related changes:

- Implements a configuration class singleton to allow the configuration settings to be accessed where needed in code instead of passing the values as function/method arguments.
- Adds Filter plugin classes for packages and releases.
- Adds functions to query package_resources for the filtering plugin classes.
- Moves the blacklist functionality to the BlacklistProject() plugin.
- Updates the mirror object blacklist functionality to use the BlacklistProject() plugin.
- Adds the new plugin entrypoints to the setup.py.
- Adds tests for the plugin functionality.
- Updated documentation for the new functionality.


src/bandersnatch/tests/test__bandersnatch_plugins__blacklist_name.py
```diff
@@ -0,0 +1,135 @@
...
+    """
+    Tests for the bandersnatch filtering classes
+    """
+    tempdir = None
+    cwd = None
+
+    def setUp(self):
+        self.cwd = os.getcwd()
+        self.tempdir = TemporaryDirectory()
+        bandersnatch.filter.loaded_filter_plugins = defaultdict(lambda: [])
```
cooperlees: What does lambda achieve?

src/bandersnatch/tests/test_mirror.py
```diff
@@ -92,13 +95,52 @@ def test_mirror_generation_4_resets_status_files(tmpdir):
...
 
 
-def test_mirror_remove_blacklisted_packages(tmpdir):
-    blacklisted_package = 'example1'
-    m = Mirror(str(tmpdir), mock.Mock(),
-               package_blacklist=[blacklisted_package])
-    m.packages_to_sync = {'example1': None, 'example2': None}
-    m._remove_blacklisted_packages()
-    assert blacklisted_package not in m.packages_to_sync
+def test_mirror__filter_packages__match(tmpdir):
```
cooperlees: Why double __ ? What does it mean?

tox.ini
```diff
@@ -3,7 +3,7 @@ envlist = py36
 
 [testenv]
 deps = -rtest-requirements.txt
-commands = pytest
+commands = pytest -vv
```
cooperlees: Sometimes I find this to much output ... But I guess it does not hurt

src/bandersnatch/tests/test__bandersnatch_plugins__blacklist_name.py
```diff
@@ -0,0 +1,135 @@
...
+        os.chdir(self.tempdir.name)
+
+    def tearDown(self):
+        if self.tempdir:
+            os.chdir(self.cwd)
+            self.tempdir.cleanup()
+            self.tempdir = None
```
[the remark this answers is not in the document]
dwighthubbard: It's defensive coding since the teardown only removes the tempdir if one was created and stored to the self.tempdir attribute.

src/bandersnatch/filter.py
```diff
@@ -0,0 +1,122 @@
+"""
+Blacklist management
+"""
+import pkg_resources
```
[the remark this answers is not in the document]
dwighthubbard: That would certainly be a good idea, but looking at the documentation for importlib_resources at https://docs.python.org/3.7/library/importlib.html#module-importlib.resources it looks like it implements only a subset of the functionality in pkg_resources and doesn't implement the functionality to iterate registered package entrypoints.

src/bandersnatch/tests/test_mirror.py
dwighthubbard: It is to differentiate between spaces and underscores since the function names being tested have underscores in their name and I like the test to include the name of the function being tested.

src/bandersnatch/configuration.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def load_configuration(self):
+        """
+        Read the configuration from the configuration files
+        """
+        config_files = [self.default_config_file]
+        if self.config_file:
+            config_files.append(self.config_file)
+        self.config = ConfigParser()
+        self.config.read(config_files)
```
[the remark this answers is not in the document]
dwighthubbard: Using a singleton makes it easier to test since the module can be reinitialized with a different configuration during the tests.

src/bandersnatch/configuration.py
```diff
@@ -0,0 +1,46 @@
+"""
+Module containing classes to access the bandersnatch configuration file
+"""
+import pkg_resources
+from configparser import ConfigParser
+
+
+class Singleton(type):  # pragma: no cover
```
cooperlees: I'll look it up, but no idea why we need type here.

src/bandersnatch/tests/test__bandersnatch_plugins__blacklist_name.py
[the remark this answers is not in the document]
cooperlees: I think cleanup() does that.

src/bandersnatch/tests/test_mirror.py
cooperlees: Is this a standard or something? Never heard of it.

PR #18, opened 2018-05-17
Bandersnatch Verify

### LINKED ISSUE #8: Add a verify/cleanup [--delete] option
Not sure on which name - Cleanup / verify

Lets walk the file system to:
a) ensure we have all the packages (e.g. like a full sync)
b) If option supplied (--delete), remove all files found that are not found in the saved JSON metadata

Also clean up non working deleting code from each run and rely on this.

- Update documentation explaining how this verify / cleanup is the only way to clean up unneeded packages now

cooperlees:
Add a new asyncio based sub command that allows verification of package files.
We can now:
- verify all json files are up to date with PyPI
- sha256 all blobs and download if they do not match JSON metadata
- Delete files that do not belong to any JSON metadata
- [DRY RUN] mode to see what we would do :D
- If the directory the file lives in is now empty, remove it too

Disable Python 3.5 tests, we no longer work in 3.5.1

Tests to come in follow up PR

(Fixes #8)


README.md
```diff
@@ -11,7 +11,7 @@ http://www.python.org/dev/peps/pep-0381/.
 The following instructions will place the bandersnatch executable in a
 virtualenv under `bandersnatch/bin/bandersnatch`.
 
-- bandersnatch **requires** `>= Python 3.5`
+- bandersnatch **requires** `>= Python 3.6`
```
dwighthubbard: I certainly would love to require Python 3.6+ however I suspect there are a number of users of bandersnatch that are using operating systems that don't provide python releases newer than 3.5.  In which case requiring 3.6 would be a breaking change.  It seems like at minimum we need to up the major number if in fact we are dropping support for 3.5.

src/bandersnatch/verify.py
```diff
@@ -0,0 +1,193 @@
...
+logger = logging.getLogger(__name__)
+
+
+def _convert_url_to_path(url):
+    return urlparse(url).path[1:]
+
+
+async def _get_latest_json(json_path, config):  # noqa: E999
+    url_parts = urlparse(config.master)
+    url = "{}://{}/pypa/{}/json".format(
```
dwighthubbard: I get a 404 going to a url with this format.  Shouldn't it be pypi not pypa, for example:
https://pypi.org/pypa/hostlists/json -> 404
https://pypi.org/pypi/hostlists/json -> 200

cooperlees: Yeah, lol! PyPA on the mind. I need to get unit tests going.

README.md
cooperlees: Yeah I was torn here, but aiohttp made the call to be >= 3.6 and I want to move there ... We can still sync in 3.5 but not verify due to aiohttp.

cooperlees:
Why is that an issue? It only walks the JSON saved locally to verify then. Your verifier should cause the JSON to not be saved to the local disk. If it is still getting saved, I feel that’s a bug on your side.

PR #22, opened 2018-05-22
Readthedocs fix

dwighthubbard:
The doc_requirements.txt needs some more tweaks to work properly with read_the_docs.  

Currently the doc build works locally using tox -e doc_build but fails when run by read_the_docs.  The particular failure occurs with newer versions of docutils and this change pins the docutils version to one that works with read_the_docs.

### DIFF
--- a/doc_requirements.txt
+++ b/doc_requirements.txt
@@ -7,5 +7,8 @@ xmlrpc2
 # We don't want to accidently get an old version of the package that is being documented
 # bandersnatch
 
+zest.releaser
+
 sphinx
 recommonmark
+docutils==0.12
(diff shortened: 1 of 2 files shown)


doc_requirements.txt
```diff
@@ -4,8 +4,9 @@ python-dateutil
...
 requests
 xmlrpc2
-# We don't want to accidently get an old version of the package that is being documented
-# bandersnatch
+
+zest.releaser
 
 sphinx
 recommonmark
+docutils==0.12
```
cooperlees: Why do you need to lock this @ 0.12? Can it be >= 0.12?

cooperlees:
Ahh - I guess we just close it?

dwighthubbard:
This was rolled up into another PR, closing

Issue #28, opened 2018-05-23
Filter plugins are not seeing configuration settings


dwighthubbard:
The filtering plugins aren't seeing configuration settings from the configuration file because the configuration singleton does not get created with the configuration file parsed from main()

PR #29, opened 2018-05-23
Filter plugin fixes

dwighthubbard:
This is to address issue #28 

The problem is the configuration object being used by the plugins does not load the configuration file passed as a command line argument in main() as a result the plugins are ignoring setting values from the actual configuration file.

This fixes the issue by populating the configuration singleton with the configuration file name.

This also includes some code that evidently didn't get included in the release plugin that was needed to actually validate the release versions.

### DIFF
--- a/src/bandersnatch_filter_plugins/blacklist_name.py
+++ b/src/bandersnatch_filter_plugins/blacklist_name.py
@@ -35,7 +35,6 @@ class BlacklistProject(FilterProjectPlugin):
         filtered_packages = set()
         try:
             lines = self.configuration['blacklist']['packages']
-            logger.debug('blacklist->packages: %r', lines)
             package_lines = lines.split('\n')
         except KeyError:
             package_lines = []
@@ -45,10 +44,6 @@ class BlacklistProject(FilterProjectPlugin):
                 continue
             package_requirement = Requirement(package_line)
             if package_requirement.specifier:
-                logger.debug(
-                    'Package line %r has a version spec, ignoring',
-                    package_line
-                )
                 continue
             if package_requirement.name != package_line:
                 logger.debug(
@@ -81,10 +76,7 @@ class BlacklistProject(FilterProjectPlugin):
             return False
 
         if name in self.blacklist_package_names:
-            logger.debug(
-                'MATCH: Package %r is in %r', name,
-                self.blacklist_package_names
-            )
+            logger.debug(f'MATCH: Package {name!r} is in the blacklist')
             return True
         return False
 
@@ -118,7 +110,6 @@ class BlacklistRelease(FilterReleasePlugin):
         filtered_requirements = set()
         try:
             lines = self.configuration['blacklist']['packages']
-            logger.debug('blacklist->packages: %r', lines)
             package_lines = lines.split('\n')
         except KeyError:
             package_lines = []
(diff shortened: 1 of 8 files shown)


src/bandersnatch_filter_plugins/blacklist_name.py
```diff
@@ -31,20 +35,15 @@ def _determine_filtered_package_names(self):
         filtered_packages = set()
         try:
             lines = self.configuration['blacklist']['packages']
-            logger.debug('blacklist->packages: %r', lines)
             package_lines = lines.split('\n')
         except KeyError:
             package_lines = []
         for package_line in package_lines:
```
cooperlees: NIT newline here is nice so it looks seperate to the try/except - Won't block on it.

src/bandersnatch_filter_plugins/blacklist_name.py
```diff
@@ -72,26 +71,55 @@ def check_match(self, **kwargs):
...
 
-        logger.info(
-            'Checking for package %s in %r', name, self.blacklist_package_names
-        )
         if name in self.blacklist_package_names:
-            logger.debug(
-                'MATCH: Package %r is in %r', name,
-                self.blacklist_package_names
-            )
+            logger.debug(f'MATCH: Package {name!r} is in the blacklist')
```
cooperlees: Isn't name a str, why do you want the __repr__ of it?

src/bandersnatch_filter_plugins/blacklist_name.py
```diff
@@ -72,26 +71,55 @@ def check_match(self, **kwargs):
...
     name = 'blacklist_release'
 
+    def initialize_plugin(self):
+        """
+        Initialize the plugin
+        """
+        # Generate a list of blacklisted packages from the configuration and
+        # store it into self.blacklist_package_names attribute so this
+        # operation doesn't end up in the fastpath.
+        self.blacklist_release_requirements = \
```
cooperlees: I prefer using () over \ for line continuation - so does the `black` code style which I will soon be converting this code base too and adding a black run into travis

src/bandersnatch_filter_plugins/blacklist_name.py
```diff
@@ -72,26 +71,55 @@ def check_match(self, **kwargs):
...
+        try:
+            lines = self.configuration['blacklist']['packages']
+            package_lines = lines.split('\n')
+        except KeyError:
+            package_lines = []
+        for package_line in package_lines:
+            package_line = package_line.strip()
+            if not package_line or package_line.startswith('#'):
+                continue
+            filtered_requirements.add(Requirement(package_line))
```
cooperlees: This parsing code looks like code above - utility function?

src/bandersnatch_filter_plugins/blacklist_name.py
```diff
@@ -111,9 +139,25 @@ def check_match(self, **kwargs):
             True if it matches, False otherwise.
         """
         name = kwargs.get('name', None)
-        version = kwargs.get('version', None)
+        version_string = kwargs.get('version', None)
 
-        if not name or not version:
+        if not name or not version_string:
             return False
```
cooperlees: maybe add a debug log for future use?

src/bandersnatch_filter_plugins/blacklist_name.py
dwighthubbard: It put's the string quotes around the name

src/bandersnatch_filter_plugins/blacklist_name.py
dwighthubbard: Unfortunetly, the ( doesn't occur in the line then until after 80 columns.

PR #33, opened 2018-05-24
Add whitelist plugin

dwighthubbard:


cooperlees:
Wait what? Why is there so many commits on this PR mate? It feels wrong.

cooperlees:
Thanks for cleaning!

PR #34, opened 2018-05-24
Improve logging of the filter operations

dwighthubbard:
Add an is_project_filtered() function to easily find if a project is in
a filter rule..

Added more debug logging to show when releases are getting filtered.

Fix a one character typo that was causing an exception in the release
filters because the wrong attribute name was being used.

Fixed a log message that was printing the formatting string because
the string wasn't an f-string.

### DIFF
--- a/src/bandersnatch/mirror.py
+++ b/src/bandersnatch/mirror.py
@@ -1,4 +1,4 @@
-from .filter import filter_project_plugins
+from .filter import is_project_filtered
 from .package import Package
 from .utils import rewrite, USER_AGENT, update_safe
 from packaging.utils import canonicalize_name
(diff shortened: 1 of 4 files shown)


src/bandersnatch/mirror.py
```diff
@@ -128,18 +128,17 @@ def _cleanup(self):
 
     def _filter_packages(self):
         """
-        Run the package filtering plugins and remove any packages from the
-        packages_to_sync that match any filters.
+        Run the package filtering plugins and remove any packages from
+        self.packages_to_sync that match any filters.
         """
         packages = list(self.packages_to_sync.keys())
```
cooperlees: Why do we need to make this a list? Can we just use the generator for the loop below?
- The deletion happens on the key in the dict, so this is not needed from what I see.

dwighthubbard: You cannot remove keys from a dictionary while iterating over them, the following code will generates an exception:
``` python
packages = dict(a=1,b=2,c=3)
packages.keys()
for package in packages:
    del packages[package]
packages.keys()
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-1-b76a4ea0ad81> in <module>()
      1 packages = dict(a=1,b=2,c=3)
      2 packages.keys()
----> 3 for package in packages:
      4     del packages[package]
      5 packages.keys()

RuntimeError: dictionary changed size during iteration
```

cooperlees:
Please reopen if you're going to do this.

Issue #15, opened 2018-05-16
PEP484 type annotate all of bandersnatch


cooperlees:
Lets add PEP484 type annotations to all source code and get clean passing `mypy` checks.

Mypy: http://mypy.readthedocs.io/en/stable/getting_started.html
typing module: https://docs.python.org/3/library/typing.html

pradyunsg:
Useful tools here:

- https://github.com/dropbox/pyannotate/
- https://github.com/Instagram/MonkeyType

cooperlees:
Ok - Only plugins left to strictly type annotate. Help welcome or I'll eventually get to it.

cooperlees:
I didn't mean to close this! lol

rkm:
Hi! I can probably finish this off if you can point me in the right direction

cooperlees:
No problem. What's left is all the test code here:

https://github.com/pypa/bandersnatch/tree/master/src/bandersnatch/tests

## How I'd go about this
- Remove `src/bandersnatch/tests/.*` from https://github.com/pypa/bandersnatch/blob/master/.pre-commit-config.yaml#L33
- Make a virtualenv - install requrements.txt + requirements_dev.txt
- run mypy and make it all pass :D

rkm:
hi, just a quick update. I've made a start on this here: https://github.com/pypa/bandersnatch/compare/master...rkm:issue/15-pep484-type-annotations. Turns out pytest is not very pep484-friendly currently, so I've had to import some of its private members to keep mypy happy. Interested to hear any suggestions!

some related issues:
- pytest typing - https://github.com/pytest-dev/pytest/issues/3342
- false unreachable after `pytest.raises` block - https://github.com/python/mypy/issues/8985

cooperlees:
Awesome stuff! Really appreciate the efforts. I'm also open to accepting some PRs that partly make this work and to divide up your remaining plan. I knew this would not be a super easy PR to finish. Also happy to accept plenty of `# type: ignores` and we can try go back through them as things gets resolved - e.g. pytest 6 arrived.

If you'd like more real time chat we can do that too on IRC - irc.freenode.net #bandersnatch. We have a few active maintainers hang out in there very regularly now!

PR #40, opened 2018-06-16
Introduce mypy and enforce it on a few files

asottile:
to be honest, I'm not exactly sure how I got here -- I was reading through the project to try and understand what it does and noticed #15 and somehow my first thought was "oh hey, that sounds like fun".  I applied it to a few modules and then got to a point where I needed to actually understand the code to make further progress and so I decided to commit what I had and make a PR.

This also (selfishly?) adds [pre-commit](https://pre-commit.com) as an easier way to run linting (~I noticed there wasn't currently any automated linting in CI so I added it!~ EDIT: had to change from pep8 -> flake8 so that type annotations weren't false positives) -- this part isn't entirely necessary as mypy can be invoked directly.

### DIFF
--- a/src/bandersnatch/configuration.py
+++ b/src/bandersnatch/configuration.py
@@ -1,13 +1,10 @@
 """
 Module containing classes to access the bandersnatch configuration file
 """
-from typing import Any
-from typing import Dict
-from typing import Optional
-from typing import Type
+from configparser import ConfigParser
+from typing import Any, Dict, Optional, Type
 
 import pkg_resources
-from configparser import ConfigParser
 
 
 class Singleton(type):  # pragma: no cover
--- a/src/bandersnatch/configuration.py
+++ b/src/bandersnatch/configuration.py
@@ -1,14 +1,19 @@
 """
 Module containing classes to access the bandersnatch configuration file
 """
+from typing import Any
+from typing import Dict
+from typing import Optional
+from typing import Type
+
 import pkg_resources
 from configparser import ConfigParser
 
 
 class Singleton(type):  # pragma: no cover
-    _instances = {}
+    _instances: Dict['Singleton', Type] = {}
 
-    def __call__(cls, *args, **kwargs):
+    def __call__(cls, *args: Any, **kwargs: Any) -> Type:
         if cls not in cls._instances:
             cls._instances[cls] = super(
                 Singleton, cls
@@ -17,7 +22,7 @@ class Singleton(type):  # pragma: no cover
 
 
 class BandersnatchConfig(metaclass=Singleton):
-    def __init__(self, config_file=None):
+    def __init__(self, config_file: Optional[str] = None) -> None:
         """
         Bandersnatch configuration class singleton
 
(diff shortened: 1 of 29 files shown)


src/bandersnatch/configuration.py
```diff
@@ -1,14 +1,19 @@
 """
 Module containing classes to access the bandersnatch configuration file
 """
+from typing import Any
+from typing import Dict
+from typing import Optional
+from typing import Type
```
cooperlees: Why not all one from?
i.e.
from typing import (Any, Dict, Optional, Type)

src/bandersnatch/utils.py
```diff
@@ -89,7 +94,7 @@ def update_safe(filename, **kw):
             prefix=os.path.basename(filename) + '.', **kw) as tf:
         if os.path.exists(filename):
             os.chmod(tf.name, os.stat(filename).st_mode & 0o7777)
-        tf.has_changed = False
+        tf.has_changed = False  # type: ignore
```
cooperlees: What's the error here? I feel we should be able to fix it.

test-requirements.txt
```diff
@@ -1,7 +1,4 @@
-flake8
```
cooperlees: Don't we still use flake8?

src/bandersnatch/configuration.py
asottile: personal preference though if you'd like a different style I will change it.

I use [reorder_python_imports](https://github.com/asottile/reorder_python_imports) -- the style here aims to reduce merge conflicts above everything else.

If you'd like, I could set you up with `isort` in this PR as well, that way we don't have to bicker about needless style

src/bandersnatch/utils.py
asottile: `has_changed` is not an attribute of file objects, there's attribute stuffing happening here

test-requirements.txt
asottile: yes, but via pre-commit.  pre-commit manages the installation in a separate environment so you don't have to worry about system installation of tools

cooperlees:
All sounds good - If you're down to add `isort` too that would be awesome. Then I'll follow up with a black style pre-commit + PR so style is behind us :)

test-requirements.txt
cooperlees: I only like people to have the ability to run things manually.
- Sorry about all the questions, I am a GitHub noob, been locked away in Facebook for to long.

I'm not quite Dave Harrington n00b, but close.

tox.ini
```diff
@@ -19,3 +19,14 @@ extras = doc_build
...
+
+[testenv:lint]
+basepython=python3
+skip_install=True
+deps=pre-commit
+commands=
+    pre-commit run --all-files --show-diff-on-failure
+
+[isort]
+include_trailing_comma = True
```
cooperlees: I'll fix this if black does not agree here. I can't remember.

cooperlees:
I'll also plan to use MonkeyType or like tool to get more type annotations for free.

test-requirements.txt
asottile: oh hey there @DaveHarrington -- small world I guess?

PR #45, opened 2018-06-18
Run yesqa against all the files

asottile:
https://github.com/asottile/yesqa

I had to [manually edit one comment][0].

- this line previously excluded any lint on that line
- `noqa` comments for a specific code [need a `:`][1] (yeah, their comment is
  wrong)
- yesqa knows this so it converted `# noqa F821` -> `# F821`

[0]: https://github.com/pypa/bandersnatch/blob/ccc9a38/src/bandersnatch/package.py#L89
[1]: https://github.com/PyCQA/flake8/blob/3.5.0/src/flake8/defaults.py#L49

### DIFF
--- a/src/bandersnatch/package.py
+++ b/src/bandersnatch/package.py
@@ -86,7 +86,7 @@ class Package:
         try:
             # If symlink already exists throw a FileExistsError
             os.symlink(self.json_file, self.json_pypi_symlink)
-        except FileExistsError:  # noqa F821
+        except FileExistsError:
             pass
 
         return True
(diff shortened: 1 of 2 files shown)


cooperlees:
So these are all things that pass now so we can remove?

asottile:
yep!

asottile:
`flake8` will read from a number of files by default, including `tox.ini` so that should (?) already work

cooperlees:
Hmmm ok, will add if I see these errors come back.

PR #48, opened 2018-06-21
bugfix/46/verify-less-memory

### LINKED ISSUE #46: Refactor verify.py to use less memory
Refactor verify.py to iterate through the file system JSON files better and use less memory.
- Potentially use a asyncio.Queue and have X (workers) consumers
-- Store only filepath in the queue and not coros

At the moment loading all the coros into a list and uses asyncio.gather is using ~1.3gb of RAM. This is not cool nor scalable.

tau3:
Hello! Please, review my attempt to fix #46

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -81,6 +81,7 @@ celerybeat-schedule
 # virtualenv
 venv/
 ENV/
+bandersnatch/
 
 # Spyder project settings
 .spyderproject
--- a/.gitignore
+++ b/.gitignore
@@ -44,6 +44,7 @@ nosetests.xml
 coverage.xml
 *,cover
 .hypothesis/
+test.conf
 
 # Translations
 *.mo
@@ -81,7 +82,6 @@ celerybeat-schedule
 # virtualenv
 venv/
 ENV/
-bandersnatch/
 
 # Spyder project settings
 .spyderproject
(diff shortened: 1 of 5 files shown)


.gitignore
```diff
@@ -44,6 +44,7 @@ nosetests.xml
 coverage.xml
 *,cover
 .hypothesis/
+test.conf
```
cooperlees: What makes test.conf? We should fix what is not cleaning it up ... can be in a follow up diff.

tau3: it's a product of `tox`

tau3:
@cooperlees 

> Please run the ‘black’ formatter over the file please.

OK, it's done. BTW, is `flake8` even enabled? I haven't seen any mentions in tox logs.

> I would like to make _VERIFIERS_COUNT configurable tho

Yep, it would be better. Should it be implemented in current or separate PR?

cooperlees:
Another recent commit turned off running flake8 via tox and changed it to precomits. I'm not sure I like it for the exact reason your hitting - Hard to run things locally.
- You're now failing mypy 

Which is also ran by travis:
https://travis-ci.org/pypa/bandersnatch/jobs/395165103

cooperlees:
Bah - I'll fix the isort.

Issue #56, opened 2018-10-16
PyPI stale caches


Xcelled:
At work, we're using a slight modified older version of Bandersnatch to mirror PyPI. About 10 days ago the mirroring job started failing - a lot - because the PyPI cache we were hitting was stale and refusing to update. After 10 days of failure, (where we got behind by about 4,000 updates), I finally forced Bandersnatch to hit a different cache by changing some request headers.

All was well for about 24 hours, then *that* cache started having stale issues. I've since updated our Bandersnatch code to use a query parameter of the current timestamp in order to unconditionally bust the caches.

That's obviously a sub-optimal solution, but it got us back up and running. It also got me thinking. instead of just bailing after a certain number of stale page errors, what if Bandersnatch fell back to using the cache-bust hack? It could retry 3 times to see if the caches get updated, and if they don't, it would force the cache to be busted in order to retrieve the proper version.

I'm willing to make a PR for this (I'm trying to upgrade our Bandersnatch at work), but I didn't want to do so without a discussion first. Maybe there's a better way to do this?

cooperlees:
Hi,

Thanks for reporting this, I have also seen this before. Let me get some other PyPA Warehouse people to read this before we rush to implement your suggestion, I also want to think about this a little more.

Cooper

Xcelled:
I can get the cache logs for you in a bit. The other modifications we made were just to upload the syncd packages to Amazon AWS. Can't imagine that's terribly useful to others.

di:
Thanks for the report! The XML-RPC API is a special snowflake and handles caching a bit differently than everything else on PyPI, so it's possible this is occasionally failing and we are unaware.

I'm a bit surprised that adding a cache breaker URL param helps here, because it shouldn't have any effect on the cached response AFAIK.

@cooperlees I'm not too familiar with Bandersnatch, can we be specific about which method would be failing here? Is it `updated_releases`?

Xcelled:
I can give a better report when I get back to my desk, but it's the call to fetch the package json that was failing. I believe the x-pypi-last-serial header was being cached improperly, as the fetched json had the new serial. Unfortunately bandersnatch checks the header.

cooperlees:
Thanks for the excellent problem description and root causing. The people I tagged above can talk about the caching better, and hopefully give us a better solution.

@di - Bandersnatch only uses the following two XMLRPC calls:
- `list_pakages_with_serial()` 
- `changelog_since_last_serial(serial: int)`
[...]

PR #57, opened 2018-10-17
Issue a purge request when we get a stale serial.

Xcelled:
Temporary fix for #56. We issue a PURGE to the url if we get a stale serial. Not ideal, but a workaround until PyPI's caches are fixed to always return the up-to-date header.

### DIFF
--- a/src/bandersnatch/master.py
+++ b/src/bandersnatch/master.py
@@ -74,7 +74,8 @@ class Master:
                     )
 
                 raise StalePage(
-                    "Expected PyPI serial {} for request {} but got {}".format(
+                    "Expected PyPI serial {} for request {} but got {}. "
+                    + "HTTP PURGE has been issued to the request url".format(
                         required_serial, path, got_serial
                     )
                 )
--- a/src/bandersnatch/master.py
+++ b/src/bandersnatch/master.py
@@ -60,6 +60,23 @@ class Master:
                         required_serial, path, got_serial
                     )
                 )
+                # HACK: The following attempts to purge the cache of the page we
+                # just tried to fetch. This works around PyPI's caches sometimes
+                # returning a stale serial for a package. Ideally, this should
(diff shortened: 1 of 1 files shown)


src/bandersnatch/master.py
```diff
@@ -60,6 +60,19 @@ def get(self, path, required_serial, **kw):
...
+                logger.debug(f"Issuing a PURGE for {path} to clear the cache")
+                try:
+                    self.session.request("PURGE", path, timeout=self.timeout)
+                except requests.exceptions.HTTPError:
+                    logger.warning(
+                        "Got an error when attempting to clear the cache", exc_info=True
+                    )
+
                 raise StalePage(
                     "Expected PyPI serial {} for request {} but got {}".format(
```
cooperlees: Can we add to the exception message that we’ve sent a PURGE (more to remind me / other devs in the future)

Xcelled: Yeah but iirc the stalepage exception is just swallowed, so the message isn't seen.

cooperlees: It will if you have `stop-on-error` set in the config file.

Xcelled: Done.

TechCiel:
Now PURGE is no more open but requiring credentials, so this HACK leads to 401 error flooding the log. Shall we consider removal of this workaround?

see also: https://github.com/pypa/bandersnatch/issues/1220

About the root cause, have a look at https://github.com/pypi/warehouse/issues/12214

cooperlees:
I did see this was removed, forgot about it. Naught me should have cut an issue. I’d be more than happy to pull it out and release a new minor version. PR welcome.

Will cleanup the code too and stop wasting peoples resources trying.

ichard26:
> [...] Naught me should have cut an issue. [...]

https://github.com/pypa/bandersnatch/issues/1334 there ya go!

Issue #59, opened 2018-10-18
verify --json-update doesn't work; incorrect calls


electricworry:
The executor parameter to url_fetch call is missing.

cooperlees:
Going to split up the PRs

PR #60, opened 2018-10-18
Fix url_fetch calls made as part of --json-update. 

electricworry:
Fixes pypa/bandersnatch#59.

### DIFF
--- a/src/bandersnatch/main.py
+++ b/src/bandersnatch/main.py
@@ -109,6 +109,11 @@ def main():
         help="Do not download or delete files",
     )
     v.add_argument(
+        "--proxy",
+        type=str,
+        help="Specify proxy server if required",
+    )
+    v.add_argument(
         "--json-update",
         action="store_true",
         default=False,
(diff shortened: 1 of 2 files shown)


src/bandersnatch/main.py
```diff
@@ -108,6 +108,11 @@ def main():
         default=False,
         help="Do not download or delete files",
     )
+    v.add_argument(
+        "--proxy",
+        type=str,
+        help="Specify proxy server if required",
```
cooperlees: Can we please give an example to make it easier for people to use. Proxy server URLs differ everywhere.

electricworry: Hello,
Certainly; I'm happy to do whatever is required! Regarding the PR, I'm a bit new to these. I originally did one for the first one (#59) but my subsequent commits have all been automatically absorbed onto the PR. Do I need to create a specific tag for each PR?

Once I know what I should be doing, I can cancel this PR and get separate PRs for each individual issue done properly. Also, I'll provide usage for proxy as requested.

Cheers!

cooperlees:
Broken this into multiple PRs:
- #71
- #72
- #74

Closing. Thanks for the work on this tho and finding the bugs! Super appreciated!

PR #64, opened 2018-10-18
Move to stable 3.7 + 3.8 travis runs

cooperlees:
- Allow 3.8 to fail for now ...


cooperlees:
As this does nothing but improve testing and it all passes. I’m merging.

PR #69, opened 2018-10-29
Update all third party dependencies

cooperlees:
- Lets sync all upstream deps and then I'll test running on a test instance on my laptop


cooperlees:
Will canary test @ Facebook to double check and have asked @dwighthubbard if he could @ Yahoo too.

PR #74, opened 2018-10-30
Set aiohttp to respect proxy vars like requests/urllib3

### LINKED ISSUE #73: Make aiohttp respect HTTP Environment Variables like Requests
We should support environment variables for aiohttp like requests.

Pass `proxy_from_env=True` to session operation(s)

cooperlees:
Set aiohttp to respect proxy vars like requests/urllib3
- Works but has a bug at the moment of printing that ~/.netrc is missing - https://pastebin.com/tSYtrvjr
- Tracked here: https://github.com/aio-libs/aiohttp/issues/3298
- Fixes #73

### DIFF
--- a/src/bandersnatch/verify.py
+++ b/src/bandersnatch/verify.py
@@ -89,8 +89,12 @@ async def verify(
         logger.debug(f"Not trying to sync package as {json_full_path} does not exist")
         return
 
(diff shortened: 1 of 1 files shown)


cooperlees:
Going to merge but won't release Bandersnatch 3.1 until we have aiohttp > 3.4.4

PR #77, opened 2018-10-30
Store N versions of index.html

### LINKED ISSUE #9: Look at keeping X old index.html files for package and main Simple API html
We should add some protection from upstream changes causing blank index.html files to be generated. For example:

https://github.com/pypa/warehouse/issues/3551 (Although this particular scenario seems intentional).
Lets come up with an archiving system that keeps X (x == int set in config file) copies of index.html so we can compare changes over time and restore if needed locally.

0 == disabled feature.

yeraydiazdiaz:
Configurable via `keep_index_versions`

Fixes #9

### DIFF
--- a/src/bandersnatch/default.conf
+++ b/src/bandersnatch/default.conf
@@ -55,6 +55,14 @@ stop-on-error = false
 ; Number of consumers which verify metadata
 verifiers = 3
 
+; Number of prior simple index.html to store. Used as a safeguard against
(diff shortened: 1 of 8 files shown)


cooperlees:
Also if you don't have time say so and I'll do the follow up diff to plumb this up from the config file.

cooperlees:
Lets aim for 3.1 for this.

Issue #23, opened 2018-05-22
Add a Windows CI Run


cooperlees:
Lets ensure we stay working on Windows.

1) Check all unit tests pass on Windows
2) Add us to appveyor or any other service that gives us Windows containers etc.

charlievil:
I just started to work on this issue during the PyPA Sprint Weekend at Bloomberg (2018). I think there is a dependency that has to be worked on module mirror.py, using fctnl not supported on windows. 
https://ci.appveyor.com/project/charlievil/bandersnatch/builds/19864677/job/tt6qv62c3dc6nlaf#L110

```
__________ ERROR collecting src/bandersnatch/tests/test_mirror.py ____________
ImportError while importing test module 'C:\projects\bandersnatch\src\bandersnatch\tests\test_mirror.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
src\bandersnatch\tests\test_mirror.py:12: in <module>
    from bandersnatch.mirror import Mirror
.tox\py\lib\site-packages\bandersnatch\mirror.py:4: in <module>
    import fcntl
E   ModuleNotFoundError: No module named 'fcntl
```

cooperlees:
Ahhh awesome - This is why we needed this!

I guess we need to look at a cross platform locking mechanism. I have very little Windows knowledge it seems we either need to use:
- oslo.concurrency
- fastners (seems less active)

Any Windows gods out there who know of the best way to get Linux / Mac / Windows `flock`ing ...

I might try take a look tonight - Please put your PR up if it's ready as I'd love the CI to show me it's fixed.

charlievil:
This is just the branch unfinished with some changes also in the setup.py(that just require some though, if this is the better way to include this change), the change tries to correct the use the module src.bandersnatch when its running the setup.py that just creates and install this module(its a module that depends on itself because of the version reference in the __init__).

This is the PR which I started to work and I am more than happy if someone continues to do real work on it :) 
https://github.com/charlievil/bandersnatch/tree/testing/23-add-appveyor-config

cooperlees:
Ok thanks - Apparently Travis can do Windows too now -So I think we should try that once the flock becomes OS aware.
https://docs.travis-ci.com/user/reference/windows/

charlievil:
@cooperlees I run the windows tests on the master branch. See the failing tests:
https://ci.appveyor.com/project/charlievil/bandersnatch/builds/20013693/job/yw4plvpkd90tlhsj

[...]

PR #82, opened 2018-11-01
Move from *nix only fnctl.flock

cooperlees:
- support Windows + other os and File Locks


cooperlees:
Helps allow #23 to happen.

PR #83, opened 2018-11-14
Pre-release filter

### LINKED ISSUE #54: Ability to filter prerelease versions
Either by specific project or wholesale. Top 20 or so disk consumers are prereleases that come out nightly for large packages.

yeraydiazdiaz:
~WIP towards addressing~ Fixes #54 

Before I move further with this I'd like to check if the implementation is valid and acceptable in the context of bandersnatch development.

AFAIK pre-releases are simply releases whose name end with `rcX`, `alphaX` or `betaX` (I might be missing more) so rather than a bespoke filter I thought it'd be more flexible to create a generic regex filter and specify different patterns. We could create a bespoke filter that is simply a preconfigured regex filter if it makes it more straight forward to use though.

While working on this I was a bit confused on a couple of things, maybe someone can guide me:

1. The plugins are imported via entry points, which was a bit of a trap for me because the tests did not recognize the new plugin even though I've updated `setup.py` until I realized that I needed to re run `pip install -e .` to pick up the changes. Is there any particular benefit from this approach? The plugins seems fairly lightweight and importing them all shouldn't be much trouble, but I might be missing something.
2. [The plugins config is always under the `blacklist` option](https://github.com/pypa/bandersnatch/blob/master/src/bandersnatch/tests/test__bandersnatch_plugins__whitelist_name.py#L69-L75), in particular blacklist/whitelist is a bit counter intuitive, unless it just happens to be how that test is set up and you can do it some other way.


cooperlees:
Will ping @dwighthubbard as he's been the lead designer here.

Since we have multiple plugins now I'd personally like a section per plugin for the config file.

src/bandersnatch/package.py
```diff
@@ -147,15 +147,15 @@ def sync(self, stop_on_error=False, attempts=3):
...
         """
-        Run the release filtering plugins and remove any packages from the
-        packages_to_sync that match any filters.
+        Run the release filtering plugins and remove any releases from
+        `releases` that match any filters.
         """
         versions = list(self.releases.keys())
         for version in versions:
-            filter = False
+            filter_ = False
```
dwighthubbard: What is the reason to change the name of this variable?

yeraydiazdiaz: just to avoid name clashing with the `filter` built-in

yeraydiazdiaz:
Thanks @dwighthubbard, I've updated the PR with your suggestions.

src/bandersnatch_filter_plugins/prerelease_name.py
```diff
@@ -0,0 +1,43 @@
+import logging
+import re
+
+from bandersnatch.filter import FilterReleasePlugin
+
+logger = logging.getLogger(__name__)
```
cooperlees: In #85 I had to change plugin logging to pull the "bandersnatch" logger in order to actually output - Are you sure these loggers are outputting? If so, I'd love to know why I didn't ...

src/bandersnatch_filter_plugins/regex_name.py
```diff
@@ -0,0 +1,89 @@
+import logging
+import re
+
+from bandersnatch.filter import FilterProjectPlugin, FilterReleasePlugin
+
+logger = logging.getLogger(__name__)
```
cooperlees: I expect same here

src/bandersnatch/tests/plugins/test_blacklist_name.py
```diff
@@ -191,3 +191,5 @@ def test__plugin__loads__default(self):
         plugins = bandersnatch.filter.filter_release_plugins()
         names = [plugin.name for plugin in plugins]
         self.assertIn("blacklist_release", names)
+
+        # TODO: add test for existing release plugin
```
cooperlees: I don't follow this - I could just be slow - What's the existing release plugin?

yeraydiazdiaz:
Hi @cooperlees, I've made the requested changes, as for the intregration test, I'm not sure how to best implement it.

At the moment we're using a single `ci.conf` with two whitelisted packages but I believe we only assert `peerme` was correctly downloaded. Bash is not exactly my strong suit but adding the prerelease filter to that single conf and asserting for all packages/releases sounds quite complicated to do in a single bash script.

My first instinct would be to try to use Python for the integration tests as well, I'm not sure of any limitations on what you can do in Travis on that front. Otherwise we could have many simple config and bash scripts testing specific elements of bandersnatch which would probably be easier to manage and extend but likely to have some repetition as bash doesn't lend itself to code reuse afaik.

cooperlees:
Bash has functions. And you can source a bash file from one to the other. I could also write it in Python, I was just keeping it simple but I might swap it to Python to make it easier for Pythonistas to contribute.

I’ll merge this and I’ll add to the CI when I get time.

I also don’t get the benefits of the plugin architecture being like this. I accepted as @dwighthubbard did the work and it works.

Issue #81, opened 2018-11-01
Pinning Deps reduces package Flexibility


cooperlees:
If you don't use a virtualenv and have large monorepos pinning deps hard can cause a lot of uneeded pain. For example at the moment, bandersnatch has everything set via `==`, in a shared repos if someone moves requests or any bandersnatch dependency forward bandersnatch will complain and fail to execute.

## Suggestions and would love some feedback:
- Change requirements to >= to at least allow moving forward and ONLY pin if a newer version causes issues - *Could introduce more travis runs when new versions come out etc.*
- Find the lowest version we work with and `>=` it in requirements.txt
- Go back to un-versioned setup.py install_requires and leave it up to the deployment to respect requirements.txt or not

dwighthubbard:
I think the unversioned setup.py install_requires would be correct.  Pinning versions in setup.py is a [bad pattern](https://packaging.python.org/discussions/install-requires-vs-requirements/) since the setup.py should be used to specify what a project minimally needs to run correctly.

If we were going to put any versioning of dependencies in the setup.py I would think pinning the minimum versions and possibly capping the major versions for some packages would be reasonable.

The requirements.txt and test-requirements.txt files should have pinned concrete dependencies to create a known working environment.

So I would be all for moving the pinned versioning out of setup.py and putting abstract deps in say setup.cfg.

yeraydiazdiaz:
I agree with @dwighthubbard, depending on how cutting edge we want to be we could consider a [`pyproject.toml`](https://www.python.org/dev/peps/pep-0518/) and/or [Poetry](https://github.com/sdispater/poetry)

cooperlees:
Agree - Reversing my setup.py change. Thanks all!

cooperlees:
Have reverted to non versioned setup.py, but it seems most people would like atleast >= in the setup.py for minimum deps. I'll work on a PR for that.

I also want to setup some autoupdate setup, like Warehouse. This will be low risk once we have a CI Integration test that is being tracked in #78

PR #84, opened 2018-11-15
Revert setup.py to specify unversioned deps

cooperlees:
- Lets allow people to install latest deps
- Leave requirements.txt that hard pin known working versions
Test install in a venv: https://pastebin.com/FUzsgcX6

Addresses #81

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,13 +6,10 @@ from setuptools import find_packages, setup
 
 from src.bandersnatch import __version__
 
-assert version_info >= (3, 6, 1), "black requires Python >=3.6.1"
+assert version_info >= (3, 6, 1), "bandersnatch requires Python >=3.6.1"
 
 
-def get_install_deps():
-    """ Load requirements.txt to get pinned versions installed """
-    with open("requirements.txt") as rfp:
-        return rfp.read().split("\n")
+INSTALL_DEPS = ("aiohttp", "packaging", "requests", "setuptools", "xmlrpc2")
 
 
 setup(
@@ -28,7 +25,7 @@ setup(
     packages=find_packages("src"),
     package_dir={"": "src"},
     include_package_data=True,
-    install_requires=get_install_deps(),
+    install_requires=INSTALL_DEPS,
     entry_points="""
             [bandersnatch_filter_plugins.project]
                 blacklist_project = bandersnatch_filter_plugins.blacklist_name:BlacklistProject
--- a/src/bandersnatch/__init__.py
+++ b/src/bandersnatch/__init__.py
@@ -19,7 +19,7 @@ __version_info__ = _VersionInfo(
     major=3,
     minor=1,
     micro=0,
-    releaselevel="dev0",
+    releaselevel="dev1",
     serial=0,  # Not currently in use with Bandersnatch versioning
 )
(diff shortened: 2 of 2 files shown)


ewdurbin:
I'm -1 on this as it will leave `pip install bandersnatch` at the whimsy of our dependencies.

I realize that moving to pinned deps is a newer thing, but is arguably the correct choice for ensuring that `pip install bandersnatch` creates a working bandersnatch installation.

ewdurbin:
Suggested alternatives:

  - Get [pyup](https://pyup.io) or [dependabot](https://dependabot.com) wired up on the repo to consistently be updating our `requirements.txt` to effectively "test" the behavior of fresh bandersnatch installations "in the field". Cut new releases as we fix compatibility.
  - Stick with pinned deps, upgrade them as often as possible. ([pyup](https://pyup.io) or [dependabot](https://dependabot.com)) could help here.
  - Build out a complex matrix of tests for varying versions of dependencies and get busy trying to maintain compatibility across them.

cooperlees:
My goal was eventually to get pyup / something to auto update and keep requirements.txt doing exactly this.

The hard pinning works great for environments that use venvs per app, or blindly pull in deps from PyPI, but not everywhere (Facebook bias here) do that. This hard pinning causes issues if shared internal built + tested versions of said dependencies are not to the exact release version pinned to.

I feel offering both a requirements.txt with the pinned versions allows people to do either. That is:
```
# People who want pinned:
pip install -r requirements.txt bandersnatch
# People who want freedom + live with the risk
pip install bandersnatch
```

yeraydiazdiaz:
Can I ask what's the main use case for bandersnatch?

To me it feels like it's an application that uses `pip` as its installation mechanism, which is a slightly awkward place to be since, as @cooperlees mentions, it almost forces the user to create a venv just for it or only install other packages that are fairly looser in their dependencies.

If that's the case I'd side with having pinned dependencies by default and having instructions as to how to install it with unpinned (or possibly `>=`) versions for the less common case the user needs to install alongside other packages.

ewdurbin:
Little bit of a broader discussion, but when I consider tools like:

- `pip` and `virtualenv`
- [`awscli`](https://github.com/aws/aws-cli/blob/8d3117aa6349bbc1a765df61d9febc0b5fd8cd7f/setup.py#L26-L31)
- [`black`](https://github.com/ambv/black/blob/e4e59f87a8c4d3844dca9a14b6a4abf544c7a0c8/setup.py#L44)
- [`tox`](https://github.com/tox-dev/tox/blob/1e8fa7ba9d88aefb1888cc0a951a252f7098f311/setup.py#L29-L37)

That have a place as general purpose utilities they choose one of two routes:

- Vendor _everything_ (`pip`, `virtualenv`)
- Have lax but well specified ranges of dependencies (`awscli`, `black`, `tox`)

None of these are "applications" per say but utilities or command line tools. This is rather similar to bandersnatch after all... it's just a script that happens to create a mirror of PyPI then keep it up to date.

Is `pip install` the best way to distribute these? Maybe not. But they should be reliable.

I'm in favor of unpinned but ranged dependencies if we're testing within that range. It's the right middle ground between pinned and chaos.

ewdurbin:
wrong button :-D

PR #87, opened 2018-11-20
Add pyup shield to README.md

cooperlees:
- Show we're up to date!


cooperlees:
Just adding a badge so abusing admin privledges.

Issue #95, opened 2018-11-26
example1 & example2 included in default blacklist


jakevdp:
I've found that if you don't specify a blacklist in ``bandersnatch.conf``, the ``bandersnatch mirror`` command acts as if you have included the example blacklist from the documentation.

here is my configuration file at ``/etc/bandersnatch.conf``:
```
[mirror]

; The directory where the mirror data will be stored.
directory = /srv/pypi

; Save JSON metadata into the web tree:
json = false

; The PyPI server which will be mirrored.
master = https://pypi.org

; The network socket timeout to use for all connections.
timeout = 10

; Number of worker threads to use for parallel downloads.
workers = 3

; Whether to hash package indexes
; Recommended setting: the default of false for full pip/pypi compatibility.
hash-index = false

; Whether to stop a sync quickly after an error is found or whether to continue
; syncing but not marking the sync as successful. Value should be "true" or
; "false".
stop-on-error = false

; Number of consumers which verify metadata
verifiers = 3
```
And here is the beginning of the output:
```
$ bandersnatch mirror
2018-11-26 22:41:07,145 INFO: bandersnatch/3.1.1 (cpython 3.6.7-final0, Linux x86_64)
2018-11-26 22:41:07,146 INFO: Setting up mirror directory: /srv/pypi/
2018-11-26 22:41:07,146 INFO: Setting up mirror directory: /srv/pypi/web/simple
2018-11-26 22:41:07,147 INFO: Setting up mirror directory: /srv/pypi/web/packages
2018-11-26 22:41:07,147 INFO: Setting up mirror directory: /srv/pypi/web/local-stats/days
2018-11-26 22:41:07,147 INFO: Generation file missing. Reinitialising status files.
2018-11-26 22:41:07,148 INFO: Status file missing. Starting over.
2018-11-26 22:41:07,148 INFO: Syncing with https://pypi.org.
2018-11-26 22:41:07,148 INFO: Current mirror serial: 0
2018-11-26 22:41:07,148 INFO: Syncing all packages.
2018-11-26 22:41:11,704 INFO: Package 'example2' is blacklisted
2018-11-26 22:41:11,704 ERROR: example2 not found in packages to sync - blacklist_project has no effect here ...
2018-11-26 22:41:11,949 INFO: Package 'example1' is blacklisted
2018-11-26 22:41:11,949 ERROR: example1 not found in packages to sync - blacklist_project has no effect here ...
...
```
[...]

PR #96, opened 2018-11-27
bug fix: reading either default or system configuration instead of both

GreatBahram:
I fixed this [issue](https://github.com/pypa/bandersnatch/issues/95). #95 
I think the problem was configuration module tries to read default and system configuration. The problem is when you comment any section in system configuration, it'll read that section through default configuration.

### DIFF
--- a/src/bandersnatch/configuration.py
+++ b/src/bandersnatch/configuration.py
@@ -37,8 +37,10 @@ class BandersnatchConfig(metaclass=Singleton):
 
     def load_configuration(self) -> None:
         """
-        Read the configuration from the configuration files
+        Read the configuration from a configuration file
         """
-        config_file = self.config_file if self.config_file else self.default_config_file
+        config_file = self.default_config_file
+        if self.config_file:
+            config_file = self.config_file
         self.config = ConfigParser()
         self.config.read(config_file)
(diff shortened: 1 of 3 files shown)


cooperlees:
Thanks for this!

Seems we need to always add an empty [blacklist] section - Please look at the failing tests and fix the code so we always atleast have an empty list. Or we will need to fix the plugin + tests to not require this.

You can run tests locally by installing and running `tox` from the root of the bandersnatch directory.

Looking deeply and my running 3.1.1 in prod, it seems this is just cosmetic and the code actually works. We could exempt these from printing the logs for example1 and example2
- calling on @dwighthubbard to suggest too please!

GreatBahram:
Check it now, please. I made the changes you said

cooperlees:
I did notice that, and I tried to clean it all up, but I got a lot of main test failures and didn't want to attack it. PRs welcome if all CI tests pass.

GreatBahram:
OK, I'll work on it. should I open an issue for that?

cooperlees:
I'm a less formal guy and this is general cleanup - Happy for no Issue.

GreatBahram:
OK, @cooperlees. The problem is like the previous one, we should clear `Singleton._instances` for each test case. the problem is, these test case are some normal function I mean they're not class that inherited from `unittest.TestCase` and I have no idea how to write setUp function without class, maybe I should put all the test cases inside a class?
other ideas?

GreatBahram:
I've just fixed that. I'll send the PR

PR #100, opened 2018-11-29
remove reading two configuration file in main.py + add changes to tes…

GreatBahram:
remove reading two configuration file in main.py


GreatBahram:
Sorry for that, these continuous integration test makes me crazy, they need more time than coding.

cooperlees:
Please run `tox` locally then! It will all tell you before you submit the diff if everything but the Integration test will pass ...

GreatBahram:
OK, I'll use that, I just tested the code with pytest and that was OK but ...

PR #102, opened 2018-11-30
add moving status file to /tmp and add it command-line option

GreatBahram:
Hi 
I've added an option to move status file to /tmp directory and force bandersnatch to check the pypi master.

### DIFF
--- a/src/bandersnatch/main.py
+++ b/src/bandersnatch/main.py
@@ -95,7 +95,8 @@ def main():
         "--force-check",
         action="store_true",
         default=False,
-        help="Force bandersnatch to check everything against the master PyPI.",
+        help="Force bandersnatch to reset the PyPI serial (move serial file to /tmp) to \
+                perform a full sync",
     )
     m.set_defaults(op="mirror")
 
@@ -165,8 +166,8 @@ def main():
                 logger.debug(
                     "Force bandersnatch to check everything against the master PyPI."
                 )
-            except Exception as e:
-                logger.debug("Could not move status file: {}".format(str(e)))
+            except OSError as e:
+                logger.error("Could not move status file: {}".format(str(e)))
         mirror(config)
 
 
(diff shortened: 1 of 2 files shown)


src/bandersnatch/main.py
```diff
@@ -152,6 +158,15 @@ def main():
...
             loop.close()
     else:
+        if args.force_check:
+            status_file = os.path.join(config.get("mirror", "directory"), "status")
+            try:
+                shutil.move(status_file, "/tmp/status")
+                logger.debug(
+                    "Force bandersnatch to check everything against the master PyPI."
+                )
+            except Exception as e:
```
cooperlees: Please don't Pokemon exception handle, (gotta catch em all) let us catch what we expect it to throw.
```suggestion
            except OSError as e:
```

src/bandersnatch/main.py
```diff
@@ -152,6 +158,15 @@ def main():
...
+                logger.debug("Could not move status file: {}".format(str(e)))
```
cooperlees: I feel if someone wants this to happen and it does not, this is an error ...
```suggestion
                logger.error("Could not move status file: {}".format(str(e)))
```

GreatBahram: You're right. The reason I chose debug is, I though these message makes the program ugly, since they appears before anything. I though maybe it's better to show them only when a user wants to debug the program.

cooperlees:
I'm going to merge but I'll make it more Windows friendly. We have a small goal to work on Windows too.

GreatBahram:
Aha, That sounds interesting. What makes it windows friendly? Just for my curiosity :-)

cooperlees:
Using the `os.path` for `pathlib` libraries + tempfile to find the respective temp location for that platform is always a preferred option. Please look at my tweaks here:
https://github.com/pypa/bandersnatch/commit/e480bded168d9191a4666640819c153dc76280a5

GreatBahram:
Thank you @cooperlees

PR #109, opened 2018-12-12
prevent executing bandersnatch when no argument has entered

GreatBahram:
Hi

Sometimes I just want to see `bandersnatch` options but it tries to mirror everything so I changed the code for that part and I added `--version` option to make sure I'm using the latest version due to `force-check` option.

thanks

### DIFF
--- a/src/bandersnatch/main.py
+++ b/src/bandersnatch/main.py
@@ -4,6 +4,7 @@ import configparser
 import logging
 import logging.config
 import shutil
+import sys
 from os import path
 from tempfile import gettempdir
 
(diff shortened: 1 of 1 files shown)


cooperlees:
This is what ‘—help’ is for ...

Changing this might break people’s cron jobs who already rely on this executing without specifying any option. Bandersnatch is a config driven process for this reason.

GreatBahram:
Actually, when i ran it without any argument I got this message:
```bash
(tmp-ecc3c8d7a6fd8a75) ➜ tmp-ecc3c8d7a6fd8a75 bandersnatch
  File "/home/bahram/.virtualenvs/tmp-ecc3c8d7a6fd8a75/bin/bandersnatch", line 11, in <module>
    sys.exit(main())
  File "/home/bahram/.virtualenvs/tmp-ecc3c8d7a6fd8a75/lib/python3.6/site-packages/bandersnatch/main.py", line 155, in main
    if args.op == "verify":
AttributeError: 'Namespace' object has no attribute 'op'
(tmp-ecc3c8d7a6fd8a75) ➜  tmp-ecc3c8d7a6fd8a75 
```

GreatBahram:
Your welcome

Issue #88, opened 2018-11-20
Automate Bandersnatch Docker Image Building


cooperlees:
Docker is the hotness. We should create a Docker image on the Docker Hub / Registry for all to use.

cooperlees:
https://github.com/pypa/bandersnatch/pull/386 Has made this happen on master.

We should now make an action to tag releases on Docker Hub on GitHub release creation.

SanketDG:
> We should now make an action to tag releases on Docker Hub on GitHub release creation.

Looks like this can be easily done with a easy combination of https://help.github.com/en/actions/reference/events-that-trigger-workflows#release-event-release and https://github.com/elgohr/Publish-Docker-Github-Action

Do you want me to go ahead and work on this?

cooperlees:
Please! Awesome.

PR #113, opened 2018-12-16
Add a reference Docker File

cooperlees:
- Use latest Python 3
- Install all tested Python module version in requirements.txt

Tested build via `docker build --network=host --tag bandersnatch .`:
https://pastebin.com/0ZGgMUtA

Addresses #88 - Will see what the best way to automate building this image on every commit or release? Don't know what's the hotness with Docker (lost inside Facebook)


GreatBahram:
Yeah that was what I meant like `python setup.py docker`, so just ignore it.

cooperlees:
Waited long enough for objections. Going to squash and merge.

PR #124, opened 2019-01-07
Make azure run tests on Mac OS X

cooperlees:
Can not hurt to test on more platforms. Next goal will be Windows on azure too.


cooperlees:
Sweet - that worked first go - We haz Mac OS X testing!

PR #149, opened 2019-01-31
Enable codecov in tox runs

cooperlees:
- Lets log our coverage over time ...


cooperlees:
All passes. Using my power to merge.

Issue #165, opened 2019-03-04
Blacklist doesn't work at Debian 9


sharoue:
At debian 9 I put both blacklist and whitelist in .conf file like this

[blacklist]
packages =
    CodeIntel
    Fiona
    FlexGet
    OpenVisus
    Panda3D
    ....(top 100 largest packages I dont want)
[whitelist]
packages =
    3to2
    51degrees-mobile-detector
    51degrees-mobile-detector-v3-wrapper
    ....(5000 popular packages I want)

but start mirroring it comes out that
2019-03-04 14:48:23,760 xxxx: Initialized project plugin 'whitelist_project', filtering ['kaptan', 'tf-estimator-nightly', ...(names of 5000 whtielisted packages)]
2019-03-04 14:48:24,038 xxxx: No project filters are enabled. Skipping filtering

and the blacklisted packages is still downloaded, blacklist doesnt work.
Is there anything wrong?

Both list worked when I run mirroring at Windows, but the NTFS filesystem not good.

cooperlees:
I'd expect bugs with Windows. #23 is still not done. Help finishing Windows support is welcome! Docker is also a good choice if you need to run it on Windows, although I'm not sure how `docker volumes` work Linux Containers to Windows underneath.

As for you lists, you just need to explicitly add in that you want to use the plugins. I.e.:
```
[blacklist]
plugins =
    blacklist_project
    whitelist_project 
...
```
- Yes I don't like that it's under blacklist, but it's legacy unfortunately :(

I've attached a PR that tries to make this clearer. I just need to learn how to push newer documentation. @dwighthubbard please help me here or tell me how I get access to do it.

cooperlees:
Documentation is clearer - Reopen if you want to discuss the windows issues.

PR #166, opened 2019-03-04
Update Documentation about being explicit with filter plugins

cooperlees:
- Update documentation (need to workout how to push it)
- Clean up some CI conf
- Rename requirements files to be more unified

Addresses #165 #142


cooperlees:
Have finally logged a task with TravisCI to see why Bandersnatch is getting flagged for abuse. I would expect it's downloading MBs from PyPI.org - Will see if I can get whitelisted for this test. Or I'll reduce the packages we pull.

Issue #68, opened 2018-10-29
Add support for `data-requires-python` attribute


zzh1996:
According to [PEP 503](https://www.python.org/dev/peps/pep-0503/)

> A repository MAY include a `data-requires-python` attribute on a file link. This exposes the `Requires-Python` metadata field, specified in PEP 345, for the corresponding release. Where this is present, installer tools SHOULD ignore the download when installing to a Python version that doesn't satisfy the requirement. For example:
> 
> `<a href="..." data-requires-python="&gt;=3">...</a>`
> In the attribute value, `<` and `>` have to be HTML encoded as `&lt;` and `&gt;`, respectively.

Without this feature, users will encounter errors because the latest version of a package may conflict with the running Python version.

e.g. using a mirror synced by bandersnatch
```
# python2 -m pip install ipython -U
Looking in indexes: https://mirrors.ustc.edu.cn/pypi/web/simple
Collecting ipython
  Downloading https://mirrors.ustc.edu.cn/pypi/web/packages/5b/e3/4b3082bd7f6908af828561b0129b5064bff4a13e6acadb321fc2d939a605/ipython-7.0.1.tar.gz (5.1MB)
    100% |################################| 5.1MB 7.7MB/s 
    Complete output from command python setup.py egg_info:
    
    IPython 7.0+ supports Python 3.5 and above.
    When using Python 2.7, please install IPython 5.x LTS Long Term Support version.
    Python 3.3 and 3.4 were supported up to IPython 6.x.
    
    See IPython `README.rst` file for more information:
    
        https://github.com/ipython/ipython/blob/master/README.rst
    
    Python sys.version_info(major=2, minor=7, micro=15, releaselevel='candidate', serial=1) detected.
    
    
    
    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-install-Tm_NkK/ipython/
```

There will be no error if you switch to the official PyPI repository since `data-requires-python` attribute is provided.

konetzed:
Any progress on this?

cooperlees:
No. It's not been a priority of mine. It seems this also requires the maintainer to actually fill in the "requires-python" `setup.py/cfg` field. E.g. https://pypi.org/pypi/peerme/json is Python 3 only in the classifiers but I (as it's my package) didn't put in requires-python cause I didn't even know about it when I made this package years ago. So I don't feel it's going to be super high signal.

PR is welcome tho. You would basically:
[...]

PR #175, opened 2019-03-24
Generate data-requires-python attributes in index.html

z4yx:
This commit solves issue #68

### DIFF
--- a/src/bandersnatch/package.py
+++ b/src/bandersnatch/package.py
@@ -1,4 +1,5 @@
 import hashlib
+import html
 import logging
 import os.path
 import sys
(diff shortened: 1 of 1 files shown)


src/bandersnatch/package.py
```diff
@@ -204,6 +205,15 @@ def sync_release_files(self):
 
         self.mirror.altered_packages[self.name] = downloaded_files
 
+    def gen_data_requires_python(self, release):
+        if "requires_python" in release and release["requires_python"] is not None:
```
cooperlees: It will never be None - Seems it will only be "" ... so just:
```
and release["requires_python"]
```
- Would probably do, but this will work.

Issue #49, opened 2018-07-12
Only download latest versions


wayneworkman:
Is there any way to configure bandersnatch to only download the latest version of packages? I don't need nor want 600GB+ of every released version of every package. I'm not running an official mirror, I just want a local copy of latest packages.

cooperlees:
Nope, not today. But I'd accept the PR.

tl;dr from memory way I'd approach it:
- make a config option
- Find the code that reads the JSON for the package
- check if latest version / checksum exists
- Delete all other versions

I also think devpi fits your needs better potentially too:
https://devpi.net/docs/devpi/devpi/stable/%2Bd/index.html

dwighthubbard:
Seems like once file filtering plugins are added we could just implement this as a plugin.

RWoodring79:
I have been tasked with setting up a PyPI mirror on a standalone network but space is somewhat limited. Is the filter plugin idea feasible yet?

cooperlees:
It does not exist, but I'd look at just whitelisting the packages you need OR blacklist the top 100 space users documented here: https://pypi.org/stats/

I even have a small tool to generate a bandersnatch blacklist: https://github.com/cooperlees/pypistats

RWoodring79:
What does not exist, the filtering infrastructure or this specific filter? 

If I understand the numbers correctly, the entire PyPI mirror would be 3.1 TB but chopping off the top 100 would cut it down to about 1 TB. I started down this path, but I do not have enough storage space for that on my internet facing machine. The simple answer is to add more storage, but sometimes learning a new technology is faster than navigating the corporate bureaucracy for IT purchasing.

cooperlees:
The logic to only keep the latest version on your mirror does not exist.

You will potentially run into problems when packages hard pin version of their deps only maintaining the latest version.

I'd prefer maybe X latest versions and it be configurable. Using that and whitelist should allow people to get pretty lean and mean mirrors.

wayneworkman:
That sounds amazing to me.

RWoodring79:
I have been looking into this and have some prototype code hacked into package.py:_filter_release(). It needs more testing before I attempt to make a filter (and a better understanding of how the filters work), but it seems feasible. 
@cooperlees Do you use and IDE (like PyCharm?) for debugging, or are you just much better at Python than me? I have something set up and working now, but it is clunky at best, debugging code under site-packages in my venv.

cooperlees:
[...]

PR #183, opened 2019-04-01
Add latest version and specific platform plugins

rene-d:
This PR brings a solution for https://github.com/pypa/bandersnatch/issues/49 and allows to remove unwanted platform-specific binaries.

Thus, two new plugins were added.

* `LatestReleaseFilter` based on `FilterReleasePlugin`. I had to change the filtering method since `check_match` is not relevant to that job. So, I propose `filter` that takes the full dictionary of releases and returns a filtered one (or the same, if the plugin is not configured). The plugin is called after the other release plugins, (see `Package._filter_latest`). I'm not very satisfied with the kind of hack I used... (`hasattr(plugin, "filter")`), if you find a better solution that not breaks the other ones.

* `ExcludePlatformFilter`, based on a new kind of plugin. The `check_match` method takes a file_desc as parameter (the dict that describes a file of a release). I have tried to be the most precise in filtering. Maybe setuptools or pip has a function that extracts the target platform (source, Windows 32/64, Linux 32/64, and so one) from the filename. Recent packages are well named, but older ones are really a mess.

The tests and linting with tox are ok (however with Python 3.7.3, I don't have a py3.6 environment anymore). 

I have added documentation also.

### DIFF
--- a/docs/filtering_configuration.md
+++ b/docs/filtering_configuration.md
@@ -129,4 +129,6 @@ plugins =
 keep = 3
 ```
 
-By default, the plugin keeps 3 releases. You should be aware that it can break requirements.
+By default, the plugin does not fiter out any release. You have to add the `keep` setting.
+
+You should be aware that it can break requirements.
--- a/docs/filtering_configuration.md
+++ b/docs/filtering_configuration.md
@@ -99,3 +99,34 @@ releases =
 ```
 
 Note the same `filter_regex` section may include a `packages` and a `releases` entry with any number of regular expressions.
+
+
+### Platform-specific binaries filtering
+
+This filter allows advanced users not interesting in Windows/macOS/Linux specific binaries to not mirror the corresponding files.
+
+
+``` ini
+[blacklist]
+plugins =
+    exclude_platform
+platforms =
+    windows
+```
+
+Avalaible platforms are: `windows` `macos` `freebsd` `linux`.
+
+
+### Keep only latest releases
+
+You can also keep only the latest releases based on greatest [Version](https://packaging.pypa.io/en/latest/version/) numbers.
+
+``` ini
+[blacklist]
+plugins =
+    latest_release
+[latest_release]
+keep = 3
+```
+
+By default, the plugin keeps 3 releases. You should be aware that it can break requirements.
--- a/src/bandersnatch_filter_plugins/filename_name.py
+++ b/src/bandersnatch_filter_plugins/filename_name.py
@@ -5,50 +5,6 @@ from bandersnatch.filter import FilterFilenamePlugin
 
 logger = logging.getLogger("bandersnatch")
 
-_plugin_initialized: bool = False
-_patterns: List[str] = []
-_packagetypes: List[str] = []
-
-
-def _init_once(self):
-    global _plugin_initialized
-
-    if _plugin_initialized:
-        return
-
-    try:
-        for platform in self.configuration["blacklist"]["platforms"].split("\n"):
-
-            lplatform = platform.lower()
-            if lplatform in ("windows", "win"):
-                _patterns.extend([".win32", "-win32", "win_amd64", "win-amd64"])
-                _packagetypes.extend(["bdist_msi", "bdist_wininst"])
-
-            elif lplatform in ("macos", "macosx"):
-                _patterns.extend(["macosx_", "macosx-"])
-                _packagetypes.extend(["bdist_dmg"])
-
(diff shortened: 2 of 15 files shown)


cooperlees:
Many thanks for the PR. I’m on leave at the moment. I will review this Wednesday mountain time.

rene-d:
Thanks :)
But I think I missed something... The "latest releases" filter should take account of the whitelist, especially if a version specifier is given.
I'll try to implement this, in a few days.

cooperlees:
We can do this as a follow up diff if you like, this diff / PR is already large.. I'll still take a look now and offer feedback.

docs/filtering_configuration.md
```diff
@@ -99,3 +99,36 @@ releases =
...
+
+``` ini
+[blacklist]
+plugins =
+    exclude_platform
+platforms =
+    windows
+```
+
+Avalaible platforms are: `windows` `macos` `freebsd` `linux`.
```
cooperlees: Should we link to and support all https://packaging.python.org/specifications/platform-compatibility-tags/ too?

src/bandersnatch_filter_plugins/filename_name.py
```diff
@@ -0,0 +1,94 @@
...
+def _init_once(self):
+    global _plugin_initialized
+
+    if _plugin_initialized:
+        return
+
+    try:
+        for platform in self.configuration["blacklist"]["platforms"].split("\n"):
+
+            if platform.lower() in ("windows", "win"):
```
cooperlees: Can you set a var and call it lplatform to ensure always lowered rahter than calling it ever if/else

```
lplatform = platform.lower
if lplatform in ("bla", "ba"):
...
```

src/bandersnatch_filter_plugins/filename_name.py
```diff
@@ -0,0 +1,94 @@
...
+                _patterns.extend(["freebsd"])
+
+            elif platform.lower() in ("linux"):
+                _patterns.extend(
+                    [
+                        "linux-i686",
+                        "linux-x86_64",
+                        "linux_armv7l",
+                        "linux-armv7l",
+                        "manylinux1_",
```
cooperlees: There are many more here I think - Is there any libary with them defined? wheel module?

src/bandersnatch_filter_plugins/latest_name.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def filter(self, releases):
+        """
+        Filter the dictionary {(release, files)}
+        """
+        if self.keep == 0:
+            return releases
+
+        versions = map(lambda v: (parse(v), v), releases.keys())
+        latest = sorted(versions)[-self.keep :]  # noqa
```
cooperlees: What's the noqa for? Can we 
```
# noqa: ERROR_NUM
e.g.
# noqa: T484
```

rene-d: Hi, thanks for all the responses :)

I had to add #noqa because [flake8](http://flake8.pycqa.org/) (and me) does not like the space between `[-self.keep` and `:]`, but [Black](https://black.readthedocs.io/) adds it... 

I'll read the other comments tomorrow.

src/bandersnatch_filter_plugins/filename_name.py
rene-d: ok.

src/bandersnatch_filter_plugins/filename_name.py
rene-d: after reading some PEP and setuptools, I think only 'manylinux2010_' and 'linux-armv6l' are missing.
I have added them in my dev branch

src/bandersnatch_filter_plugins/latest_name.py
rene-d: changed for # noqa: E203

I think it's a bug in Black because `[-keep:]` is not reformated into `[-keep :]`. Or I miss something :-(

rene-d:
@cooperlees : do you want I submit a new PR ? I don't know how to update this one with last changes :/

docs/filtering_configuration.md
rene-d: response below

rene-d:
@cooperlees : do you prefer that I submit a new PR ?

cooperlees:
No. This all looks good - Lets merge and let some people play with it. I'm going to try and get a few more fixes in this week and release a new version. No promises tho.

Issue #196, opened 2019-04-12
3.3.0 Latest releases plugin could remove "official" version


rene-d:
In some cases (about 2500...), the most recent releases are not the official (or stable?) one referenced in the "info" section of the JSON.

I'm writing a fix.

rene-d:
BTW, I think the `verify` command will download the whole mirror, despite this filter (and platform one). I'll check and propose a fix too.

cooperlees:
Yes, verify has never been made plugin respectful. Should probably open a separate issue for that.

Verify is all async and I want to one day (once xmlrpc is dead) refactor main bandersnatch to be fully async too.

rene-d:
Well, PR submitted.

I'm still circumspect about this feature (keeping the last release)... However, I was able to reinstall all my favorite packages only with the reduced mirror. And its size is divided by 12.

PR #198, opened 2019-04-12
Fixes for https://github.com/pypa/bandersnatch/issues/196

rene-d:
Sorry one PR... read your message too late

* `verify.py` applies filters like Package to prevent downloading unwanted files
* latest releases filter always keeps the official version

### DIFF
--- a/src/bandersnatch/package.py
+++ b/src/bandersnatch/package.py
@@ -108,20 +108,22 @@ class Package:
                         package_info = self.mirror.master.get(
                             f"/pypi/{self.name}/json", self.serial
                         )
+                        package_info = package_info.json()
                     except requests.HTTPError as e:
                         if e.response.status_code == 404:
                             logger.info(f"{self.name} no longer exists on PyPI")
                             return
                         raise
 
-                    self.releases = package_info.json()["releases"]
+                    version = package_info.get("info", {}).get("version")
+                    self.releases = package_info["releases"]
 
                     self._filter_releases()
-                    self._filter_latest()
+                    self._filter_latest(version)
                     self._filter_filenames()
 
                     if self.mirror.json_save and not self.json_saved:
-                        self.json_saved = self.save_json_metadata(package_info.json())
+                        self.json_saved = self.save_json_metadata(package_info)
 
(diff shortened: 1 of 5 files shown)


cooperlees:
Many thanks for this!

I'm sorry - But I've gone and done a pretty large PR (#197) that I think we should land and then re-apply this diff on. Thoughts?

Also, could verify just include the class from `src/bandersnatch/filter.py` and use that somehow rather than duplicating the logic? (I haven't looked to hard into it)

rene-d:
Yes ok. I'll split the patch, and rebase my dev branch when your PR will be merged. 

And you're right, I have almost copy/paste some logic from Package.py to verify.py. I'm not very proud, but I didn't see how to do without refactoring filters and package.py.

rene-d:
The main idea is to replace FilterReleasePlugin and FilterFilenamePlugin (the new one I introduced for my needs - that was an error...) to a new kind of plugin that takes the whole JSON data as parameter.

So, just two kinds of filters:
* FilterProjectPlugin -> applied on the list of packages (whitelist/blacklist)
* FilterMetadataPlugin -> applied on the metadata of a package, which can remove releases and/or dist files on various criteria.

cooperlees:
Ok - I've merged my diff! Let me know if I can help. I could look at later this weekend using the `filter.py` logic in verify.

cooperlees:
I’m looking to cut 3.3.1 over the weekend. Are you going to have time to polish this up (all good if not - this can go in 3.3.2)?

rene-d:
I don't know yet, maybe tonight :)

rene-d:
Hello @cooperlees, I prefer to take my time to propose more consistent changes. FilterReleasePlugin filters could take the whole metadata. Up to the plugin to remove or keep releases and/or filenames from this.

cooperlees:
No problems! Take your time. Thanks for the PRs. I’ll cut 3.3.1 later today. I want to roll the nicer logging into my production tomorrow :)

rene-d:
Hello @cooperlees 
I will submit another PR with those changes.
We can cancel this one.

PR #216, opened 2019-05-06
Do a fresh `pip freeze` to clean up requirements.txt

cooperlees:
- Seems there are some deps we no longer need
- CI to confirm


requirements.txt
```diff
@@ -1,22 +1,19 @@
-aiodns==2.0.0
```
cooperlees: This is an extra now. need to workout how to get setup.py to want this.

PR #227, opened 2019-05-08
pyup.io: No update urllib3

cooperlees:
- Lets pin it for now so we don't keep getting updates


cooperlees:
I made a mess here somehow. Will cleanup.

PR #235, opened 2019-05-14
Version plugin api

dwighthubbard:
This is to address an issue where the plugin class filter method had different arguments in different versions of bandersnatch.   This creates an issue where bandersnatch could load plugins from a different package and which had a plugin class with a filter method with arguments different than what bandersnatch was passing.  Which would generate an exception.

This change adds a version value to the pkg_resource group for the plugins.

This allows the value to be changed when a breaking change is made to the plugin base classes.  In order to insure the plugin classes match the expected classes used by the code calling the plugins.

This change sets the version of the plugin interface on master to v2 and updates the values in the setup.cfg file to match.  As a result, bandersnatch now loads the included bandersnatch_filter_plugins from the v2 pkg_resources group.

So the included plugins continue to work correctly and external plugin packages can provide the old plugins that are compatible with the current pypi 3.3.1 release of bandersnatch using the pre versioned resource group names as well as providing and other plugin classes registered with the v2 resource group that use the newer filter method.

Since this required changes to the setup.py I migrated to using a setup.cfg for the settings since it is parseable.

The [bandersnatch_safety_db](https://github.com/dwighthubbard/bandersnatch_safety_db) package is now configured this way as an example of the changes.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -11,6 +11,9 @@ long_description_content_type = text/markdown
 license = Academic Free License, version 3
 license_file = LICENSE
 name = bandersnatch
+project_urls =
+    Source Code = https://github.com/pypa/bandersnatch
+    Change Log = https://github.com/pypa/bandersnatch/CHANGES.md
 url = https://github.com/pypa/bandersnatch/
 version = 3.4.0
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,89 @@
+[metadata]
+author = Christian Theune
+author_email = ct@flyingcircus.io
+classifiers =
+    Programming Language :: Python :: 3 :: Only
+    Programming Language :: Python :: 3.6
+    Programming Language :: Python :: 3.7
+description = Mirroring tool that implements the client (mirror) side of PEP 381
+long_description = file:README.md
+long_description_content_type = text/markdown
+license = Academic Free License, version 3
+license_file = LICENSE
+name = bandersnatch
+url = https://github.com/pypa/bandersnatch/
+version = 3.4.0
+
+[options]
+include_package_data = True
+install_requires =
+    aiohttp
+    filelock
+    packaging
+    requests
+    setuptools>40.0.0
+    xmlrpc2
+package_dir =
+    =src
+packages = find:
+python_requires = >=3.6
+
+[options.packages.find]
+where=src
+
+[options.entry_points]
+# This entrypoint group must match the value of bandersnatch.filter.PROJECT_PLUGIN_RESOURCE_GROUP
+bandersnatch_filter_plugins.v2.project =
+    blacklist_project = bandersnatch_filter_plugins.blacklist_name:BlacklistProject
+    whitelist_project = bandersnatch_filter_plugins.whitelist_name:WhitelistProject
+    regex_project = bandersnatch_filter_plugins.regex_name:RegexProjectFilter
+
+# This entrypoint group must match the value of bandersnatch.filter.RELEASE_PLUGIN_RESOURCE_GROUP
+bandersnatch_filter_plugins.v2.release =
+    blacklist_release = bandersnatch_filter_plugins.blacklist_name:BlacklistRelease
+    prerelease_release = bandersnatch_filter_plugins.prerelease_name:PreReleaseFilter
+    regex_release = bandersnatch_filter_plugins.regex_name:RegexReleaseFilter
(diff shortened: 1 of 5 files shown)


setup.cfg
```diff
@@ -0,0 +1,89 @@
+[metadata]
+author = Christian Theune
+author_email = ct@flyingcircus.io
+classifiers =
+    Programming Language :: Python :: 3 :: Only
+    Programming Language :: Python :: 3.6
+    Programming Language :: Python :: 3.7
+description = Mirroring tool that implements the client (mirror) side of PEP 381
+long_description = file:README.md
```
cooperlees: How do we get CHANGES.md now? It's nice to have the change log on https://pypi.org/p/bandersnatch
- It's not critical - Wonder if there is a way

setup.cfg
```diff
@@ -0,0 +1,89 @@
...
+version = 3.4.0
+
+[options]
+include_package_data = True
+install_requires =
+    aiohttp
+    filelock
+    packaging
+    requests
+    setuptools>40.0.0
```
cooperlees: Is > 40 needed for your stuff now? We never had that requirement before.

setup.cfg
```diff
@@ -0,0 +1,89 @@
...
+    blacklist_release = bandersnatch_filter_plugins.blacklist_name:BlacklistRelease
+    prerelease_release = bandersnatch_filter_plugins.prerelease_name:PreReleaseFilter
+    regex_release = bandersnatch_filter_plugins.regex_name:RegexReleaseFilter
+    latest_release = bandersnatch_filter_plugins.latest_name:LatestReleaseFilter
+    exclude_platform = bandersnatch_filter_plugins.filename_name:ExcludePlatformFilter
+
+console_scripts =
+    bandersnatch = bandersnatch.main:main
+
+zc.buildout =
```
cooperlees: We can prob get ird of all the zc + zest config. No one active uses it.

setup.cfg
```diff
@@ -0,0 +1,89 @@
...
+
+zest.releaser.releaser.after =
+    update_stable_tag = bandersnatch.release:update_stable_tag
+
+zest.releaser.postreleaser.after =
+    update_requirements = bandersnatch.release:update_requirements
+
+[options.extras_require]
+safety_db =
+    bandersnatch_safety_db
```
cooperlees: https://pypi.org/project/bandersnatch_safety_db/ does not exist ... Where is this ment to come from?

src/bandersnatch/main.py
```diff
@@ -15,11 +15,16 @@
 import bandersnatch.verify
 
 from .configuration import BandersnatchConfig
+from .filter import filter_project_plugins, filter_release_plugins
 
-logger = logging.getLogger(__name__)
+logger = logging.getLogger(__name__)  # pylint: disable=C0103
```
cooperlees: Why did this start firing? I also don't get why it's a `invalid-name` ... You got any insight here?

src/bandersnatch/main.py
```diff
@@ -15,11 +15,16 @@
...
 
 
 def mirror(config):
+    # Load the filter plugins so the loading doesn't happen in the fast path
```
cooperlees: What is the fast path? Do you mean, lets load them here so we hit the fast path during bandersnatch runs?

setup.cfg
dwighthubbard: I have been tempted to do that many times.

setup.cfg
dwighthubbard: https://pypi.org/project/bandersnatch-safety-db/ is published now.

src/bandersnatch/main.py
dwighthubbard: With default pylint configuration it triggers an invalid name because it wants constants to be upper case.

src/bandersnatch/main.py
dwighthubbard: Yes, the load is slow so we want them loaded before the mirroring operation starts.  

Without this change, each worker will run the plugin load.  Which slows things down a bit.  But also runs the plugin initialization.

In the case of this plugin since it fetches safety_db from github during it's initialization it also results in it running the db download from github multiple times (since each worker load the plugins)

setup.cfg
dwighthubbard: I added it under project_urls in the setup.cfg, that will add the link to the project_links section of the pypi page at https://pypi.org/project/bandersnatch/

setup.cfg
dwighthubbard: I've removed it

PR #267, opened 2019-06-14
Match prerelease versions with multiple digit suffixes

indrat:
Extend PRERELEASE_PATTERNS to match multiple digits after the prerelease tag.
- Some packages have versions beyond 9 e.g. https://pypi.org/project/scikit-vis/0.0.2.dev12/  
- Some packages use a date scheme as the suffix e.g. https://pypi.org/project/lalsuite/6.57.1.dev20190613/

### DIFF
--- a/src/bandersnatch_filter_plugins/prerelease_name.py
+++ b/src/bandersnatch_filter_plugins/prerelease_name.py
@@ -12,7 +12,7 @@ class PreReleaseFilter(FilterReleasePlugin):
     """
 
     name = "prerelease_release"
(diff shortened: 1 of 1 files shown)


cooperlees:
https://www.python.org/dev/peps/pep-0440/#id24 - PEP440 says N ... I don't see clear specification that N is one digit or not. But in reality, we prob want to filter these. Happy to accept this based on that.

I will fix the black error. In future please format your code with black - Running tox locally should identify this before CI.

Issue #293, opened 2019-07-25
Docs Failing to Build due to docutils


cooperlees:
Seems `docutils` does not have a Python 3 `0.15.1` version: https://pypi.org/project/docutils/#files
- But there is a `.post1` .tar.gz

Will try workout some solution, docutils is not on GitHub to discuss best solution.

```
$ python test_runner.py
GLOB sdist-make: /home/travis/build/pypa/bandersnatch/setup.py
doc_build create: /home/travis/build/pypa/bandersnatch/.tox/doc_build
doc_build installdeps: -rrequirements_docs.txt, sphinx-rtd-theme
ERROR: invocation failed (exit code 1), logfile: /home/travis/build/pypa/bandersnatch/.tox/doc_build/log/doc_build-1.log
================================== log start ===================================
Collecting sphinx-rtd-theme
  Using cached https://files.pythonhosted.org/packages/60/b4/4df37087a1d36755e3a3bfd2a30263f358d2dea21938240fa02313d45f51/sphinx_rtd_theme-0.4.3-py2.py3-none-any.whl
Collecting docutils==0.15.1 (from -r requirements_docs.txt (line 1))
  Could not find a version that satisfies the requirement docutils==0.15.1 (from -r requirements_docs.txt (line 1)) (from versions: 0.3, 0.3.5, 0.3.7, 0.3.9, 0.4, 0.5, 0.6, 0.7, 0.8, 0.8.1, 0.9, 0.9.1, 0.10, 0.11, 0.12, 0.13.1, 0.14rc1, 0.14rc2, 0.14, 0.15, 0.15.1.post1)
No matching distribution found for docutils==0.15.1 (from -r requirements_docs.txt (line 1))
```

cooperlees:
PR Fixed the build of Docs.

PR #294, opened 2019-07-25
Move to `docutils==0.15.1.post1`

cooperlees:
- 0.15.1 Py3 version/wheel + sdist seem gone ...

Addresses #293


cooperlees:
This works. Landing.

PR #313, opened 2019-09-11
Add a differential file list capability

artagel:
This PR adds a differential file list capability to bandersnatch.
The capability is useful in cases where a user wants to synchronize pypi to an offline server without having to determine the diffs each time a mirror command is executed.  The diff file can be used to only send diff to a synchronization disk, or to send new files through a diode to an offline or upstream pypi mirror.

### DIFF
--- a/docs/mirror_configuration.md
+++ b/docs/mirror_configuration.md
@@ -134,3 +134,28 @@ Example:
 [mirror]
 root_uri = https://example.com
 ```
+
+
+### diff-file
+
+The diff file is a string containing the filename to log the files that were downloaded during the mirror.
+This file can then be used to synchronize external disks or send the files through some other mechanism to offline systems.
+
+Example:
+```ini
+[mirror]
+diff-file = /srv/pypi/mirrored-files
+```
+
+
+
+### diff-append-epoch
+
+The diff append epoch is a boolean (true/false) setting that indicates if the diff-file should be appended with the current epoch time. 
+This can be used to track diffs over time so the diff file doesn't get cobbered each run.  It is only used when diff-file is used.
+
+Example:
+```ini
+[mirror]
+diff-append-epoch = true
+```
--- a/docs/mirror_configuration.md
+++ b/docs/mirror_configuration.md
@@ -151,7 +151,7 @@ diff-file = /srv/pypi/mirrored-files
 
 ### diff-append-epoch
 
-The diff append epoch is a boolean (true/false) setting that indicates if the diff-file should be appended with the current epoch time. 
+The diff append epoch is a boolean (true/false) setting that indicates if the diff-file should be appended with the current epoch time.
 This can be used to track diffs over time so the diff file doesn't get cobbered each run.  It is only used when diff-file is used.
 
 Example:
--- a/docs/mirror_configuration.md
+++ b/docs/mirror_configuration.md
@@ -151,7 +151,7 @@ diff-file = /srv/pypi/mirrored-files
 
 ### diff-append-epoch
 
-The diff append epoch is a boolean (true/false) setting that indicates if the diff-file should be appended with the current epoch time.
+The diff append epoch is a boolean (true/false) setting that indicates if the diff-file should be appended with the current epoch time. 
 This can be used to track diffs over time so the diff file doesn't get cobbered each run.  It is only used when diff-file is used.
 
 Example:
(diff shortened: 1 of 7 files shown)


artagel:
Not ure why that test failed, passes for me:

```
  lint: commands succeeded
  py37: commands succeeded
  congratulations :)
```

```
Trim Trailing Whitespace.................................................Passed
Fix End of Files.........................................................Passed
Check Yaml...............................................................Passed
Check for added large files..............................................Passed
Flake8...................................................................Passed
black....................................................................Passed
pyupgrade................................................................Passed
seed isort known_third_party.............................................Passed
isort....................................................................Passed
mypy.....................................................................Passed
```

cooperlees:
Hi,

I'll need to go through this one. In future can you start an issue and propose your feature rather than just turn up with a PR that could save review time. One goal, is to keep bandersnatch simple, and just syncing mirrors. Aside tho, I do like the feature this provides. It does not super complicate the code and also is a nice practice to save PyPI + WAN bandwidth.

You're running the lints in Python 3.7, travis is running in 3.6 (as we want to support it). I'm not sure why the hooks are not detecting the whitespace and formatting errors on your box (do you have environment or . (dot) files changing settings to the tools potentially?). But going into Travis output it's pretty clear what it's unhappy with here: https://travis-ci.org/pypa/bandersnatch/jobs/583730372

You're missing some black formatting + end of file whitespace. Feel free to wait for my review before you fix if you wish.

artagel:
No worries, take your time.  Test should pass now, I wasn't testing the branch :)  
Sure, I'll open an issue next time.  This is quite a useful change, especially for anyone doing offline dev.  It is quite troublesome to manually track changed files to synchronize offline pypi servers, and you are right, the code change is quite minimal. 
I made the change for our systems, so I figured I'd share it.

docs/mirror_configuration.md
```diff
@@ -134,3 +134,28 @@ Example:
 [mirror]
 root_uri = https://example.com
 ```
+
+
+### diff-file
+
+The diff file is a string containing the filename to log the files that were downloaded during the mirror.
+This file can then be used to synchronize external disks or send the files through some other mechanism to offline systems.
```
cooperlees: Can we update this with some examples. I'm guessing you're using rsync or some standard tool here? Noting that would be helpful to potential future users I feel.

src/bandersnatch/mirror.py
```diff
@@ -70,7 +74,11 @@ def __init__(
         digest_name=None,
         root_uri=None,
         keep_index_versions=0,
+        diff_file=None,
+        diff_append_epoch=False,
+        diff_full_path=None,
         flock_timeout=1,
+        diff_file_list=[],
```
cooperlees: Please do not set an immutable default (sorry should have a lint for this). It's set / evaluated @ definition time, not on function call.

http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments

```
diff_file_list=None,
```

src/bandersnatch/mirror.py
```diff
@@ -79,9 +87,13 @@ def __init__(
         self.json_save = json_save
         self.hash_index = hash_index
         self.root_uri = root_uri
+        self.diff_file = diff_file
+        self.diff_append_epoch = diff_append_epoch
+        self.diff_full_path = diff_full_path
         self.keep_index_versions = keep_index_versions
         self.digest_name = digest_name if digest_name else "sha256"
         self.workers = workers
+        self.diff_file_list = []
```
cooperlees: Might want to use your argument. If you set it to none, and it is none, it will be a new empty list with the following suggestion:

```
self.diff_file_list = diff_file_list or []
```

artagel:
I've made the changes. If there's another place you would like to add an assert for the diff_file_list please point me to it.

artagel:
> I’ll then eventually refactor some testing and move some of this logic out of main.py. It’s starting to get a little to complicated.
I have to admit it was a bit confusing following the tests.

src/bandersnatch/unittest.conf
```diff
@@ -63,6 +63,15 @@ verifiers = 3
 ; If unset defaults to 0.
 ; keep_index_versions = 0
 
+; Configure a file to write out the list of files downloaded during the mirror.
+; This is useful for situations when mirroring to offline systems where a process
+; is required to only sync new files to the upstream mirror.
+; The file be be named as set in the diff-file, and overwritten unless the
+; diff-append-epoch setting is set to true.  If this is true, the epoch date will
+; be appended to the filename (i.e. /path/to/diff-1568129735)
+diff-file = /srv/pypi/mirrored-files
```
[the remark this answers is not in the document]
artagel: Tests are passing now.

cooperlees:
Documentation is updated: https://bandersnatch.readthedocs.io/en/latest/

PR #316, opened 2019-09-14
Add Nightly Python build to CI

cooperlees:
- Lets see if `trunk` segmentation faults as well like 3.8 is at the moment


cooperlees:
Merging, this shows we don't break in latest 3.8 / nightly. Reported bug to CPython: https://bugs.python.org/issue38172

PR #323, opened 2019-09-23
[refactor] Move mirror logic out of main.py into mirror.py

cooperlees:
- Planning to add another sub command and main.py is to long
- Lets move the mirror logic to mirror.py
- Lets also ensure we always return an int and error when we should ...

Make Unit tests and CI pass to ensure we're good.

### DIFF
--- a/src/bandersnatch/main.py
+++ b/src/bandersnatch/main.py
@@ -1,118 +1,22 @@
 import argparse
 import asyncio
-import configparser
 import logging
 import logging.config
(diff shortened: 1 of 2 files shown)


cooperlees:
I expect coverage complaints ...

cooperlees:
No logic change other than some typing and a if statement bug found by mypy. CI all passes so going to merge so I can work on adding `bandersnatch delete`

PR #338, opened 2019-10-27
Added subparsers action type

### LINKED ISSUE #325: Fix `Any` typing for argparse.ArgumentParser in main.py
Line 20 we're using `Any` as mypy was complaining that argparse.ArgumentParser does not have a `add_parser` method.

sukhbeersingh:
Fix #325 : Replaced subparser function parameter from Any to argparse._SubParsersAction

### DIFF
--- a/requirements_test.txt
+++ b/requirements_test.txt
@@ -4,7 +4,7 @@ coverage==4.5.4
 flake8==3.7.8
 flake8-bugbear==19.8.0
 freezegun==0.3.12
-pre-commit==1.18.3
+pre-commit==1.19.0
 pytest==5.2.2
 pytest-timeout==1.3.3
 pytest-cache==1.0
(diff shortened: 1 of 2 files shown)


sukhbeersingh:
The official documentation only says that this is a special object type which is the only one that has access to `add_parser()` function.

> The add_subparsers() method is normally called with no arguments and returns a special action object. This object has a single method, add_parser(), which takes a command name and any
[Argparse official docs](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.add_subparsers)

Also, 

> Unfortunately the public API does not allow to get the subparsers, it only lets you add them, and do that once. So you would have to toss the subparsers object around to add something later
https://pythonhosted.org/argh/reference.html

So, no I don't think there is a public base object that we can refer to.

cooperlees:
Thanks for the excellent research and your first contribution! Will merge away.

I guess we could open a low pri argument parser bug to make it more typing friendly. I’ll think about it. Modules will become more typing friendly now it’s becoming more prevalent in code.

PR #352, opened 2019-11-12
Add GitHub Actions

cooperlees:
- Run lint, unittests + Integration
- MacOS + Linux


cooperlees:
Wow, seems Actions has found a legit issue / potential bug (didn't expect this). Will dig in later.

PR #381, opened 2019-12-25
f-string remaining .format() via flynt

cooperlees:


cooperlees:
Don't understand the coverage change :\ So going to land.

PR #391, opened 2020-01-03
Add metadata and python version filters.

TemptorSent:
First attempts at adding ability to filter packages by metadata and releases by python version.


src/bandersnatch/configuration.py
```diff
@@ -83,7 +83,8 @@ def load_configuration(self) -> None:
         config_file = self.default_config_file
         if self.config_file:
             config_file = self.config_file
-        self.config = ConfigParser()
+        self.config = ConfigParser(delimiters=('='))
```
cooperlees: Black unfortunately prefers "" ... I just roll with it cause i love auto formatters.

src/bandersnatch/package.py
```diff
@@ -152,6 +156,25 @@ def sync(self, stop_on_error=False, attempts=3):
             logger.error("Exiting early after error.")
             sys.exit(1)
 
+
+    def _filter_metadata(self, metadata):
```
cooperlees: Small nit: Type annotate since it's new code :)

```python
def _filter_metadata(self, metadata: Dict) -> bool:
```

src/bandersnatch_filter_plugins/metadata_filter.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+
+class RegexMetadataFilter(FilterMetadataPlugin):
+    """
+    Plugin to download only packages having metadata matching at least one of the  specified patterns.
+    """
+
+    name = "regex_metadata"
+    #patterns: List[Pattern] = [] # by default, keep 'em all
```
cooperlees: delete?

src/bandersnatch_filter_plugins/metadata_filter.py
```diff
@@ -0,0 +1,73 @@
...
+            config = self.configuration["regex_metadata"]
+        except KeyError:
+            return
+        else:
+            logger.info(f"Initializing regex_metadata plugin")
+            if not self.initilized:
+                for k in config:
+                    pattern_strings = [pattern for pattern in config[k].split("\n") if pattern]
+                    self.patterns[k] = [
+                        re.compile(pattern_string) for pattern_string in pattern_strings
```
cooperlees: Love the compile!

src/bandersnatch_filter_plugins/metadata_filter.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+        #for version in list(releases.keys()):
+        #release = releases[version]
+        #    if "requires_python" in release and release["requires_python"] is not None:
+        #        if any(pattern.match(release["requires_python"]) for pattern in self.patterns):
+        #            continue
+        #        else:
+        #            del releases[version]
+
```
cooperlees: Delete?

src/bandersnatch_filter_plugins/python_name.py
```diff
@@ -0,0 +1,81 @@
...
+            if "python_version" in release and release["python_version"] is not None:
+                if release["python_version"] is "source":
+                    continue
+                elif self.specifier.contains(release["python_version"]) is not None:
+                    continue
+                else:
+                    del releases[version]
+
+
+
```
cooperlees: delete

src/bandersnatch/filter.py
```diff
@@ -188,3 +214,15 @@ def filter_release_plugins() -> Iterable[Filter]:
         List of objects derived from the bandersnatch.filter.Filter class
     """
     return load_filter_plugins(RELEASE_PLUGIN_RESOURCE)
+
+
+def filter_release_file_plugins() -> Iterable[Filter]:
```
cooperlees: Probably want Sequence[Filter] here as a str is iterable ...
If interested, types are based on: https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes

src/bandersnatch/package.py
```diff
@@ -183,12 +187,46 @@ def _filter_releases(self):
...
+            return matched
+
+    def _filter_all_releases_files(self):
+        """
+        Filter release files and remove empty releases after doing so.
+        """
+        releases = list(self.releases.keys())
+        for release in releases:
+            release_files = list(self.releases[release])
+            for rfindex in reversed(range(len(release_files))):
```
cooperlees: Could save some time by only making a list once:
```python
release_files = sort(self.releases[release], reverse=True)
for rfindex in range(len(release_files))
```

cooperlees:
Ok, I've agreed with @TemptorSent that I will merge this and fix the CI Lint. I will then add tests. I will merge this when I have time to do the follow promises some time this week.

cooperlees:
tox is indeed passing, but it seems you're plugin is effecting the run of bandersnatch CI does. I went to go look but  debug output prints out every package that gets filtered, and I got sick of waiting for it to load.

We need to understand why integration fails and fix the tests or the plugins before we merge this please! Thanks.

Issue #393, opened 2020-01-04
Problems with 'runner.py'


GreatBahram:
Hello there,

I noticed these problems in the `runner.py` while I was using the docker image:

https://github.com/pypa/bandersnatch/blob/1b573a141fd6ee4b171c612b1422089b119e544f/src/runner.py#L27

* Firstly, args doesn't have a `conf` attribute, it has been defined as `config`.
* Secondly, I think, It's not a good idea to hardcode the `bandersnatch` path. Because even in the Docker image you'll find the `bandersnatch` inside `/usr/local/bin/`

* Thirdly, the `interval` type should be `int` not `str`:
https://github.com/pypa/bandersnatch/blob/1b573a141fd6ee4b171c612b1422089b119e544f/src/runner.py#L21

* Because, this line will show an error if you execute it:
https://github.com/pypa/bandersnatch/blob/1b573a141fd6ee4b171c612b1422089b119e544f/src/runner.py#L28

Even when I solved the above problems in my container, I got another error. My question is, why don't you guys utilize the default configuration when the configuration hasn't been given?
```bash
2020-01-04 16:00:45,222 WARNING: Config file '/conf/bandersnatch.conf' missing, creating default config.
2020-01-04 16:00:45,222 WARNING: Please review the config file, then run 'bandersnatch' again.
2020-01-04 16:00:45,222 ERROR: Could not create config file: [Errno 2] No such file or directory: '/usr/local/lib/python3.8/site-packages/bandersnatch/default.conf'
Sleeping for 3599.6514344215393s

```

cooperlees:
Hi! Thanks for reporting. I’ll eventually get to fixing this, but a PR is welcome anytime. Some clowny mistakes there.

P.s. I did docker to learn it - being at Facebook I hardly get to use docker.

GreatBahram:
Hi there, 

I tried to solve them, but I reckon we could achieve the configuration problem by adding some changes to Dockefile.

Let me know if there is any problem with the pull request.

cooperlees:
Suggest away to make the Docker file better :D You're the first verbose user so happy to solve your needs :)

cooperlees:
Pushed https://github.com/pypa/bandersnatch/commit/2ba069d5e7f57ce47b011d7fcbef5e68be160e09 to do everything but the described configuration problem.

I'd still prefer people bind mount in their config file to the Docker container.

GreatBahram:
[...]

PR #399, opened 2020-01-07
bugfix in runner.py:

GreatBahram:
Hello there,

Maybe this could solve some of the problems (#393 ):

* Change type of interval argument to int
* Remove hardcoded the bandersnatch path; instead run it by using the `bandersnatch.main` module

I couldn't solve the configuration part, but I think that could be achievable through Dockerfile, maybe?.

### DIFF
--- a/src/runner.py
+++ b/src/runner.py
@@ -4,27 +4,36 @@
 
 import argparse
 import sys
-from subprocess import run
+from subprocess import run, CalledProcessError
 from time import sleep, time
 
 
 def main() -> None:
-    parser = argparse.ArgumentParser(
-        formatter_class=argparse.ArgumentDefaultsHelpFormatter
-    )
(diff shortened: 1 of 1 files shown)


src/runner.py
```diff
@@ -4,27 +4,36 @@
...
+                sys.executable,
+                "-m",
+                "bandersnatch.main",
+                "--config",
+                args.config,
+                "mirror",
+            ]
+            run(cmd, check=True)
+        except CalledProcessError:
+            sys.exit(1)
```
cooperlees: You wouldn't prefer the exception to see the return value from bandersnatch itself? I feel this masks a potential more helpful return code always returning 1 ....

GreatBahram: Yeah, indeed, that would be better

cooperlees:
Will merge and fix now.

PR #404, opened 2020-01-13
Replace xmlrpc2 with aiohttp-xmlrpc

cooperlees:
- Async def up Master PyPI XML RPC Calls
- Wrap calls in `asyncio.run_until_complete` in `mirror.py`
- Pull in asynctest + use CoroutingMock's until we're >= 3.8

See integration tests still pass

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,5 @@
 aiohttp==3.6.2
+aiohttp-xmlrpc==0.7.6
 async-timeout==3.0.1
 attrs==19.3.0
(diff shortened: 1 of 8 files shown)


cooperlees:
Note: This is step 1 in removing requests from bandersnatch's dependency list

cooperlees:
Seems we need some changes to `aiohttp-xmlrpc` to fully clean up the aiohttp closing errors. Have opened an issue:
https://github.com/mosquito/aiohttp-xmlrpc/issues/27

PR #405, opened 2020-01-13
Update the development docs

yeraydiazdiaz:
Add documentation link to README


yeraydiazdiaz:
I also considered removing it entirely and linking to https://bandersnatch.readthedocs.io/en/latest/CONTRIBUTING.html but went for the least aggressive approach 🙂

PR #445, opened 2020-03-04
WIP: Add Storage Abstraction

### LINKED ISSUE #437: Abstract Storage and allow Any backend storage
We have a whole mirror for public use , it costs a lot of storage and bandwidth, so we decided to move it to s3 storage .

My current solution is to use bandersnatch sync pypi.org and then use s3cmd to sync all files to s3. But it has been tons of files , it really cusome time to do so(comparing each of files, transfer and check), I wonder if I can make use of bandersnatch , can I use bandersnatch to write files directly to s3. Or can I use the metadata in bandersnatch , like which package has been updated lately, so I can only sync those packages.

techalchemy:
- Add base class and plugins for storage abstraction layer
- Add reusable methods to the storage base class
- Add swift dependencies to `setup.cfg`
- Add a filesystem plugin for the new storage plugin system
- Provide all extant functionality via `Storage` base class
- Add `SwiftPath` Path abstraction for filesystem-like operations
- Stage temporary files to local disk before moving them to swift
- Pass all non-transient IO through the new plugin interface
- Clean up unused imports, update annotations
- Set filesystem plugin as default
- Closes #437 
- TODO: Tests, asyncio

Asyncio may be a serious constraint here because I don't know if the various object stores (s3, swift) provide asyncio libraries.  As discussed we can kick this off in a bunch of worker threads and make it pseudo-async, but it's not the same. The overhead of maintaining these independently is likely going to be pretty high and I doubt anyone wants to go that route

The real design question I am thinking about is -- do we want to keep interacting with this as though it's a filesystem, or do we want to take the opportunity to abstract away the storage backend and simply ask the `Storage` provider to supply or store a package? E.g. `storage_plugin.save_package(name, version, file)` or something? It depends how much you would want to push into the abstraction but it _could_ make sense.


src/bandersnatch/storage.py
```diff
@@ -0,0 +1,284 @@
...
+    Type,
+    Union,
+)
+
+import pkg_resources
+from packaging.utils import canonicalize_name
+
+from .configuration import BandersnatchConfig
+
+PATH_TYPES = Union[pathlib.Path, str]
```
cooperlees: https://github.com/python/typing/issues/402 - Seems one day we might have this. Could add comment. No big deal.

src/bandersnatch/verify.py
```diff
@@ -9,7 +9,7 @@
 from functools import partial
 from pathlib import Path
 from sys import stderr
-from typing import List, Set  # noqa: F401
+from typing import List, Set, Type  # noqa: F401
```
cooperlees: Why is this needed? Confused.

src/bandersnatch_storage_plugins/filesystem.py
```diff
@@ -0,0 +1,248 @@
...
+import shutil
+import tempfile
+from typing import IO, Any, Dict, Generator, List, Optional, Type, Union, overload
+
+from bandersnatch.storage import StoragePlugin
+
+logger = logging.getLogger("bandersnatch")
+
+
+PATH_TYPES = Union[pathlib.Path, str]
```
cooperlees: Import from storage?

src/bandersnatch_storage_plugins/filesystem.py
```diff
@@ -0,0 +1,248 @@
...
+    @overload
+    @contextlib.contextmanager
+    def open_file(  # noqa
+        self, path: PATH_TYPES, text: bool = True, encoding: str = "utf-8"
+    ):
+        ...
+
+    @contextlib.contextmanager
+    def open_file(  # noqa
+        self, path: PATH_TYPES, text: bool = True, encoding: str = "utf-8"
```
cooperlees: Why is this needed? Cause the args have defaults - Can't you just have the 1 definition? What's the mypy error?

src/bandersnatch/package.py
```diff
@@ -335,7 +334,9 @@ def sync_simple_page(self):
             if not self.simple_directory.exists():
                 self.simple_directory.mkdir(parents=True)
             simple_page = self.simple_directory / "index.html"
-            with utils.rewrite(simple_page, "w", encoding="utf-8") as f:
+            with self.mirror.storage_backend.rewrite(
```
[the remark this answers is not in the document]
techalchemy: This one I agree with tbh, I spent a small amount of time considering whether to dodge the extra `getattr` traversal across `.mirror` here... I'm still not too sure about it and I think it deserves more consideration.

src/bandersnatch/storage.py
techalchemy: good suggestion, let me see if this works with the inheritance i set up

src/bandersnatch/verify.py
techalchemy: ah it's probably not anymore

src/bandersnatch/storage.py
techalchemy: so `PathLike` doesn't have the `name` attribute among other things, unfortunately :|

src/bandersnatch_storage_plugins/filesystem.py
techalchemy: there is no error now apparently, i think i sorted it out :)

src/bandersnatch/tests/plugins/test_storage_plugins.py
```diff
@@ -586,6 +586,8 @@ class BaseStoragePluginTestCase(BasePluginTestCase):
 web{0}simple{0}index.html""".format(
         os.sep
     ).strip()
+    if os.name == "nt":
+        base_find_contents = base_find_contents.replace(".lock\n", "")
```
cooperlees, line 590: FWIW mypy prefers `sys.platform == "win32"`

src/bandersnatch/main.py
```diff
@@ -173,6 +177,7 @@ def main(loop: Optional[asyncio.AbstractEventLoop] = None) -> int:
 
     # TODO: Go to asyncio.run() when >= 3.7
     loop = loop or asyncio.get_event_loop()
+    loop.set_debug(True)
```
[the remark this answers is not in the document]
techalchemy: im sure I deleted this one at some point anyhow, or at least thought about it :|

src/bandersnatch/tests/plugins/test_storage_plugins.py
techalchemy: sigh. Is there any difference

cooperlees:
Easily one of the largest PRs I've ever seen for bandersnatch commit wise. I'll go over it more to make not of things to forward fix.

Then I'll merge this and change the version to 4.1.0.dev0.

techalchemy:
> Easily one of the largest PRs I've ever seen for bandersnatch commit wise. I'll go over it more to make not of things to forward fix.
> 
> Then I'll merge this and change the version to 4.1.0.dev0.

I'll likely squash/rebase in some meaningful way since a lot of them are pointless noise

src/bandersnatch/tests/ci-swift.conf
```diff
@@ -0,0 +1,25 @@
...
+enabled =
+    whitelist_project
+
+[whitelist]
+packages =
+    black
+    peerme
+    pyaib
+
+; vim: set ft=cfg:
```
cooperlees, line 25: Seem this does nothing yet? You been manually using I'd bet?

src/bandersnatch/tests/plugins/swift_container.json
```diff
@@ -0,0 +1,8829 @@
+[
```
cooperlees, line 1: Should this file not be in the PR? Looks like state for your tests?

src/bandersnatch/package.py
```diff
@@ -382,11 +388,13 @@ def _save_simple_page_version(self, simple_page_content: str) -> None:
...
-        try:
+        versions_path = (
+            self.mirror.storage_backend.PATH_BACKEND(self.simple_directory) / "versions"
+        )
+        if not versions_path.exists():
             versions_path.mkdir()
-        except FileExistsError:
-            version_files = sorted(os.listdir(versions_path))
+        else:
+            version_files = list(sorted(versions_path.iterdir()))
```
cooperlees, line 397: Huh - This shouldn't need to be wrapped in a list - Sorted returns a list ...
```python
>>> from pathlib import Path
>>> a = Path("/tmp")
>>> b = sorted(a.iterdir())
>>> type(b)
<class 'list'>
```

requirements.txt
```diff
@@ -6,10 +6,13 @@ chardet==3.0.4
 filelock==3.0.12
 idna==2.9
 importlib_resources==1.5.0; python_version < '3.7'
+keystoneauth1==4.0.0
 lxml==4.5.0
 multidict==4.7.5
+openstackclient==4.0.0
 packaging==20.3
 pyparsing==2.4.7
+python-swiftclient==3.9.0
```
cooperlees, line 15: Putting these in here make them no optional - Maybe in a separate requirements.txt?

cooperlees:
P.s. I'm trying swift.py all to you ...

PR #455, opened 2020-03-16
Remove non all PEP503 conformant paths

### LINKED ISSUE #138: Look at removing old Pip 1.5 + 8.x hacks in package.py
We should potentially look at removing the normalized_simple_directory workaround for older pips in `package.py`

cooperlees:
- PEP503 says we only need to store the normalized path / name for packages
- Moderns pip supports this
- Lets remove the legacy normalization and 1.5 non normalized directories for simple html
- Also make package.py full mypy / PEP484 compliant

Have made a note in documentation about this.

Tests:
- Update tests to only expect normalized directories

Fixes #138

### DIFF
--- a/src/bandersnatch/package.py
+++ b/src/bandersnatch/package.py
@@ -7,10 +7,9 @@ import sys
 from datetime import datetime
 from json import dump
 from pathlib import Path
-from typing import Dict, Optional
+from typing import TYPE_CHECKING, Dict, List, Optional
 from urllib.parse import unquote, urlparse
 
-import pkg_resources
 from aiohttp import ClientResponseError
 from packaging.utils import canonicalize_name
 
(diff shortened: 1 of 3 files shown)


src/bandersnatch/package.py
```diff
@@ -64,28 +66,18 @@ def normalized_simple_directory(self) -> Path:
             )
         return self.mirror.webdir / "simple" / self.normalized_name
```
techalchemy: so the difference here is the legacy format used a parent directory with the first letter as the top level directory name? I just want to be sure I understand correctly what is going away

cooperlees: As we discussed on IRC - Just the PEP's normalization function rather than some arbitrary one from back in the day in distutils.

This will also cause less to be on disk.

Issue #459, opened 2020-03-17
Renaming issue on Windows 10 (and maybe others?)


FaustinCarter:
When attempting to mirror packages on Windows 10, I get the following error, 

```
File "%localappdata%\continuum\anaconda3\lib\site-packages\bandersnatch\utils.py", line 97, in rewrite
    os.rename(filepath_tmp, filepath)
FileExistsError: [WinError 183] Cannot create a file when that file already exists: 'path\\to\\pypi_mirror\\web\\simple\\.index.html.4r6ndpfp' -> 'pypi_mirror\\web\\simple\\index.html'
```

My guess is that it is related to this issue on stack overflow: https://stackoverflow.com/questions/8107352/force-overwrite-in-os-rename

The short version being that on Windows, `os.rename` will not overwrite an existing file (unlike on Unix-based systems).

Suggested fix is to attempt to delete the target file (if exists) before attempting to overwrite it in every place where `os.rename` is called.

cooperlees:
Windows support has never been finished or declared as something bandersnatch has. Would love help in completing this. PR welcome.

https://github.com/pypa/bandersnatch/issues/23 has more history as to what I and others have tried.

cooperlees:
So, I think we are pretty good here ... Anything else left to do?

FaustinCarter:
I think this can be closed. There is definitely still an issue surrounding two of the tests that has something to do with Windows not properly deleting a lock file, but that really deserves its own issue and I can work on that separately.

cooperlees:
I think that’s the behavior of filelock library - if not please open an issue. I removed expecting a .lock file from Windows Unittestst as I thought that was expected.

FaustinCarter:
Ah, excellent. That explains why tox passes all tests now! I'm not sure what the expected behavior is, but bandersnatch is working well now and no longer has any issues on my local machine.

We should probably mention in the readme somewhere that on Windows bandersnatch must be run from an elevated prompt. I'm happy to take a stab at it if you don't want to.

cooperlees:
Feel free to draft it up or I will before I release 4.0 to PyPI.

PR #469, opened 2020-03-25
Windows

FaustinCarter:
This pull request adds some Windows support and tries to fix some tests that otherwise fail when run on a Windows 10 system. Related to issue #459 
 * A lot of fixing paths to use the right separator
 * Ignoring a test of chmod (which doesn't work on Windows)
 * Replacing os.rename with shutil.move or else Windows freaks if it has to overwrite an existing file
 * Moves a file open operation into a 'with' context block to ensure it gets closed if an exception is thrown. This allows that file to later be unlinked without Windows complaining.

This doesn't solve all the failed tests, but if I run the tests in an elevated command prompt (so that symlinking works) I've whittled it down from 20 failures to 4 failures, and those failures don't seem to impact the functionality (too much). They are:

```
======================================================================== short test summary info ========================================================================
FAILED src/bandersnatch/tests/test_mirror.py::test_mirror_empty_resume_from_todo_list - AssertionError: assert '.lock\ngener...e\\index.html' == 'generation\n...e\\ind...
FAILED src/bandersnatch/tests/test_mirror.py::test_mirror_sync_package_error_no_early_exit - AssertionError: assert '.lock\ngener...e\\index.html' == 'generation\n...e...
FAILED src/bandersnatch/tests/test_package.py::test_cleanup_non_pep_503_paths - AssertionError: assert 0 == 1
FAILED src/bandersnatch/tests/test_utils.py::test_hash - AssertionError: assert '91ef8f60d130...7543b34bfb372' == '125765989403...cb48fa3e87ff8'
```

There are still some issues however. When I test this on my system, everything works, but bandersnatch errors out with:

```
Exception ignored in: <function _ProactorBasePipeTransport.__del__ at 0x0000018276D2FF70>                                      
  File "C:\Users\fwcarter\AppData\Local\Continuum\anaconda3\envs\btest\lib\asyncio\proactor_events.py", line 116, in __del__   
    self.close()                                                                                                               
  File "C:\Users\fwcarter\AppData\Local\Continuum\anaconda3\envs\btest\lib\asyncio\proactor_events.py", line 108, in close     
    self._loop.call_soon(self._call_connection_lost, None)                                                                     
  File "C:\Users\fwcarter\AppData\Local\Continuum\anaconda3\envs\btest\lib\asyncio\base_events.py", line 715, in call_soon     
    self._check_closed()                                                                                                       
  File "C:\Users\fwcarter\AppData\Local\Continuum\anaconda3\envs\btest\lib\asyncio\base_events.py", line 508, in _check_closed 
    raise RuntimeError('Event loop is closed')                                                                                 
RuntimeError: Event loop is closed                                                                                             
```

Also, it looks like some of the builds aren't passing for one reason or another.

In any case, this doesn't fix all the problems, but bandersnatch is at least useable on my system now and it's a step in the right direction.

### DIFF
--- a/src/bandersnatch/mirror.py
+++ b/src/bandersnatch/mirror.py
@@ -124,11 +124,12 @@ class Mirror:
         processing."""
         if self.todolist.exists():
             try:
-                saved_todo = iter(open(self.todolist, encoding="utf-8"))
-                int(next(saved_todo).strip())
-                for line in saved_todo:
-                    _, serial = line.strip().split()
-                    int(serial)
+                with open(self.todolist, encoding="utf-8") as f:
+                    saved_todo = iter(f)
+                    int(next(saved_todo).strip())
+                    for line in saved_todo:
+                        _, serial = line.strip().split()
+                        int(serial)
             except (StopIteration, ValueError):
                 # The todo list was inconsistent. This may happen if we get
                 # killed e.g. by the timeout wrapper. Just remove it - we'll
--- a/src/bandersnatch/package.py
+++ b/src/bandersnatch/package.py
@@ -30,10 +30,6 @@ if TYPE_CHECKING:  # pragma: no cover
 display_filter_log = True
 logger = logging.getLogger(__name__)
 
-# Bool to help with some Windows-related path operations
-if sys.platform == "win32":
-    is_windows = True
-
 
 class Package:
 
--- a/src/bandersnatch/package.py
+++ b/src/bandersnatch/package.py
@@ -4,7 +4,6 @@ import html
 import logging
 import os.path
 import sys
-from datetime import datetime
 from json import dump
 from pathlib import Path
 from shutil import rmtree
@@ -30,6 +29,10 @@ if TYPE_CHECKING:  # pragma: no cover
 # Bool to help us not spam the logs with certain log messages
 display_filter_log = True
 logger = logging.getLogger(__name__)
+
+# Bool to help with some Windows-related path operations
+if sys.platform == "win32":
+    is_windows = True
 
 
 class Package:
@@ -366,7 +369,7 @@ class Package:
 
     def _save_simple_page_version(self, simple_page_content: str) -> None:
         versions_path = self._prepare_versions_path()
-        timestamp = datetime.utcnow().isoformat() + "Z"
+        timestamp = utils.make_time_stamp()
         version_file_name = f"index_{self.serial}_{timestamp}.html"
         full_version_path = versions_path / version_file_name
         with utils.rewrite(full_version_path, "w", encoding="utf-8") as f:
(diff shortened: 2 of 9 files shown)


FaustinCarter:
Regarding the Exception Ignored error I'm getting, I suspect it is related to this bug, which does not yet have a fix: https://bugs.python.org/issue39232.

This PR will not fix that, but it doesn't seem to cause any actual problems since the error occurs after bandersnatch successfully completes all its tasks.

src/bandersnatch/package.py
```diff
@@ -31,6 +30,10 @@
 display_filter_log = True
 logger = logging.getLogger(__name__)
 
+# Bool to help with some Windows-related path operations
+if sys.platform == "win32":
+    is_windows = True
```
cooperlees: is_windows = bool(sys.platform == "win32")

src/bandersnatch/tests/test_utils.py
```diff
@@ -64,7 +65,11 @@ def test_rewrite(tmpdir, monkeypatch):
         f.write("csdf")
     assert open("sample").read() == "csdf"
     mode = os.stat("sample").st_mode
-    assert oct(mode) == "0o100644"
+    # chmod doesn't work on windows machines. Permissions are pinned at 666
+    if sys.platform == "win32":
+        assert True
+    else:
+        assert oct(mode) == "0o100644"
```
cooperlees: All you need here is
```python
if not sys.platform == "win32":
    assert oct(mode) == "0o100644"
```

src/bandersnatch/utils.py
```diff
@@ -95,7 +103,7 @@ def rewrite(
         # put in place actually but also doesn't want to error out.
         return
     os.chmod(filepath_tmp, 0o100644)
-    os.rename(filepath_tmp, filepath)
+    shutil.move(filepath_tmp, filepath)
```
cooperlees: Nice - checked - Under the covers for POSIX systems os.rename is used ...

src/bandersnatch/mirror.py
```diff
@@ -124,11 +124,12 @@ def _cleanup(self):
         processing."""
         if self.todolist.exists():
             try:
-                saved_todo = iter(open(self.todolist, encoding="utf-8"))
-                int(next(saved_todo).strip())
-                for line in saved_todo:
-                    _, serial = line.strip().split()
-                    int(serial)
+                with open(self.todolist, encoding="utf-8") as f:
```
cooperlees: Thanks - This is much cleaner!

src/bandersnatch/package.py
FaustinCarter: Done.

src/bandersnatch/tests/test_utils.py
FaustinCarter: Done. Using new utils.WINDOWS for the check.

Issue #470, opened 2020-03-26
Bandersnatch raises exception when syncing pypi package with mixed case


FaustinCarter:
When trying to sync a package that is listed on PyPI with a non-canonical name (i.e. mixed case), bandersnatch will error out with the following exception. Example, the `pyGTC` package:

```
2020-03-26 10:28:49,379 ERROR: Error syncing package: pygtc@3330176
Traceback (most recent call last):
  File "c:\users\fwcarter\desktop\code\bandersnatch\src\bandersnatch\package.py", line 173, in sync
    self.mirror.record_finished_package(self.name)
  File "c:\users\fwcarter\desktop\code\bandersnatch\src\bandersnatch\mirror.py", line 255, in record_finished_package
    del self.packages_to_sync[name]
KeyError: 'pygtc'
```

The problem can be solved by changing `self.name` to `self.raw_name` in the following line of code.

https://github.com/pypa/bandersnatch/blob/b3f6701bcfc2fec833e50404d86b0dc96e241646/src/bandersnatch/package.py#L170

However, a better approach might be to canonicalize the names that are used as dict keys before they get used as dict keys here:

https://github.com/pypa/bandersnatch/blob/b3f6701bcfc2fec833e50404d86b0dc96e241646/src/bandersnatch/mirror.py#L195-L196

A corollary to this bug is that if one tries to whitelist a package with the canoncialized name, rather than the raw name in PyPI, then bandersnatch will fail even to find the package on PyPI. I have not tracked this down yet, but it should be considered part of this issue.

cooperlees:
Thanks for finding these bugs! I knew I'd miss something and the PR for this was just recently done by myself.

Simplicity says normalizing as early as possible (e.g. on external I/O) is the best thing to do. I will try and get a PR up to try doing this in the next few days. We will just have to live with the generated HTML having only the normalized name and forget about the raw_name concept.

For the plugin, we will just have to state we can only filter based via PEP normalized names.

Please correct anything if I am wrong.

cooperlees:
Actually, for the plugin, we should just also normalize on initialization - Then humans can put in non normalized and they will still work.

cooperlees:
Updated docs to show Case is important for filters for now. 4.1 might get the big refactor. Needed to get a release out as way to much has changed.

https://github.com/pypa/bandersnatch/pull/474

Issue #453, opened 2020-03-12
Fix test_package_sync_does_not_touch_existing_local_file to occasionally failing


cooperlees:
This is randomly failing, need to workout why. Could be a race, could be a number of things.

stat should not differ but sometimes is. Lets see what part of stat differs - It's hard to repro locally.

Example Fail: https://github.com/pypa/bandersnatch/pull/451/checks?check_run_id=504352574

cooperlees:
Another failure. Seems to always only be >= 3.8: https://github.com/pypa/bandersnatch/pull/460/checks?check_run_id=517003207

Will try get 3.8 on my Mac and try repro later today. All local dev I've been using 3.7

cooperlees:
This is still occurring on CI - I'd love a second set of eye to take a look at this to try help find my bug!

SanketDG:
Is there any consistent way of making this fail on CI?

Once it fails for a certain push, does it fail for a second push in a PR?

If that is the case, then we can use the [`stat`](https://docs.python.org/3/library/stat.html) module to inspect what exactly is different.

SanketDG:
Opened https://github.com/pypa/bandersnatch/pull/487 for debugging this.

From https://travis-ci.org/github/pypa/bandersnatch/jobs/674697261:

```
OLD STAT: os.stat_result(st_mode=33204, st_ino=2562282, st_dev=2049, st_nlink=1, st_uid=2000, st_gid=2000, st_size=0, st_atime=1586837095, st_mtime=1586837095, st_ctime=1586837095)
NEW STAT:os.stat_result(st_mode=33204, st_ino=2562282, st_dev=2049, st_nlink=1, st_uid=2000, st_gid=2000, st_size=0, st_atime=1586837096, st_mtime=1586837095, st_ctime=1586837095)
```

Looks like the access time differs by a second, can we mock this?

PR #473, opened 2020-03-27
Fix deletion of finished Packages

cooperlees:
- XMLRPC returns the raw_name
- Keeping raw_name to store this in the Simple HTML

Tests:
- Added new test `test_package_sync_handles_non_pep_503_in_packages_to_sync` that failed with using Foo as a package to sync
  - This test does not use all normal PyTest fixtures on purpose
- Added more asserts around Mirror.errors
- Returned Foo from mocked out XMLRPC return
- Fixed a few tests with other little bugs uncovered by `assert not mirror.errors`
- Added `ACMPlus` to CI Integration run

Addresses part of #470
- Still need to workout if we want to normalize all the plugin names and compare the XMLRPC returned lists with their normalized named to avoid typos etc.

### DIFF
--- a/src/bandersnatch/tests/test_package.py
+++ b/src/bandersnatch/tests/test_package.py
@@ -1,12 +1,13 @@
 import os.path
 import unittest.mock as mock
 from pathlib import Path
-from tempfile import gettempdir
+from tempfile import TemporaryDirectory, gettempdir
 from typing import List
 
 import pytest
 from freezegun import freeze_time
 
+from bandersnatch.mirror import Mirror
 from bandersnatch.package import Package
 from bandersnatch.utils import make_time_stamp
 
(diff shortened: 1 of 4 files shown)


src/bandersnatch/tests/test_package.py
```diff
@@ -397,22 +404,34 @@ def test_save_json_metadata(mirror, package_json):
...
+
 @pytest.mark.asyncio
 async def test_package_sync_does_not_touch_existing_local_file(mirror):
     pkg_file_path_str = "web/packages/any/f/foo/foo.zip"
     pkg_file_path = Path(pkg_file_path_str)
     touch_files((pkg_file_path,))
-    with pkg_file_path.open("wb") as f:
-        f.write(b"")
+    with pkg_file_path.open("w") as f:
+        f.write("")
```
cooperlees: This might also help with #453

PR #480, opened 2020-04-04
Add unittest for new utils.safe_name

cooperlees:
- Add missing dependencies for importlib_resources from #479
- Rename utils.safe_name to bandersnatch_safe_name
- Also now pre-compile the safe_name regex
- Add unittest for bandersnatch_safe_name


SanketDG:
Looks good to me @cooperlees

PR #482, opened 2020-04-05
Add GH Action for pushing to Docker Hub on Release

SanketDG:
Added the Github Action for automating pushing the `pypa/bandersnatch`
image to Docker Hub when a new release in published.

Closes https://github.com/pypa/bandersnatch/issues/88

### DIFF
--- a/.github/workflows/docker_upload.yml
+++ b/.github/workflows/docker_upload.yml
@@ -17,3 +17,4 @@ jobs:
           username: ${{ secrets.DOCKER_USERNAME }}
           password: ${{ secrets.DOCKER_PASSWORD }}
           snapshot: true
+          tag_names: true


SanketDG:
I see.

Would using We can always use https://github.com/elgohr/Publish-Docker-Github-Action/blob/master/README.md#tag_names to tag the images work out?

cooperlees:
Yup - Seems all we need is `tag_names: true` in .github/workflows/docker_upload.yml. Thanks for finding that!

cooperlees:
There might be a race before the commit gets the release tag ...
- I wonder if it's re-ran when a tag is applied?

SanketDG:
> I wonder if it's re-ran when a tag is applied?

Interesting, let me test this out.

SanketDG:
This is blocked for the meantime, my local Docker instance seems to be causing issues.

SanketDG:
Pulled this from the graves! Apologies, I get what I deserve for distrohopping so frequently!
I have updated the changes to use `tag_names` for releases.

> There might be a race before the commit gets the release tag ...
> 
> * I wonder if it's re-ran when a tag is applied?

So I tested this, and it seems like it only tags the image when an *annotated* tag (which is a tag + commit) is pushed, so it only creates an image with the tag name.

PR #487, opened 2020-04-14
Fix test_package_sync_does_not_touch_existing_local_file to occasionally failing

SanketDG:


src/bandersnatch/tests/test_package.py
```diff
@@ -434,7 +434,9 @@ async def test_package_sync_does_not_touch_existing_local_file(mirror):
     assert not mirror.errors
 
     # Use Pathlib + create a new object to ensure no caching
-    assert old_stat == Path(pkg_file_path_str).stat()
+    # Only compare the relevant stat fields
+    assert old_stat.st_mtime == Path(pkg_file_path_str).stat().st_mtime
+    assert old_stat.st_ctime == Path(pkg_file_path_str).stat().st_ctime
```
cooperlees, line 439: Personally I would have performed 1 stat, saved to a variable and asserted on the returned tuple. Here you’re doing two calls to disk, rather than 1.

PR #492, opened 2020-04-23
The first option to fix problem described in aiohttp-xmlrpc/pull/35

mosquito:


cooperlees:
Thanks for the effort. I get I could make this work, but I feel my PR to your library will make it much friendly to use. Especially the next PR I have planned where I planned to add `async def __aenter__()` + `async def __aexit__()` and then the lines of code in bandersnatch will be a lot less!

Ideally I'd like to not need to make my own context managers and use one our of your library like how I use aiohttp.

Also, I'd recommend using tox locally to run tests: 
https://bandersnatch.readthedocs.io/en/latest/CONTRIBUTING.html#running-unit-tests

That said, I'm not sold as these PRs are cleaner than letting me make your library API friendlier.

cooperlees:
Went with https://github.com/pypa/bandersnatch/pull/506

Issue #494, opened 2020-04-24
deleted packages on pypi cause sync failure


windtail:
This maybe a bug of aiohttp, but i am not familiar with aiohttp.

The following lines in `Master::get` function 

```python
        async with self.session.get(path, timeout=timeout, **kw) as r:
            got_serial = (
                int(r.headers[PYPI_SERIAL_HEADER])
                if PYPI_SERIAL_HEADER in r.headers
                else None
            )
            await self.check_for_stale_cache(path, required_serial, got_serial)
            yield r
```

if `r.status == 404`, `aiohttp.ClientResponseError` is **NOT** raised, instead `StalePage` is raised later, so the condition of deleted packages is not detected correctly

windtail:
The following hack may fix this
```patch
Index: src/bandersnatch/master.py
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- src/bandersnatch/master.py	(revision 3fcdc9fd0722fd06bc0ecfdda3704371a57f5ae8)
+++ src/bandersnatch/master.py	(revision 7418eb2f05491e8d484291174d0f9c9e0065eb52)
@@ -104,6 +104,8 @@
             del kw["timeout"]
 
         async with self.session.get(path, timeout=timeout, **kw) as r:
+            if r.status != 200:
+                raise aiohttp.ClientResponseError(r.request_info, r.history, status=r.status)
             got_serial = (
                 int(r.headers[PYPI_SERIAL_HEADER])
                 if PYPI_SERIAL_HEADER in r.headers
```

cooperlees:
Hi,
Thanks for the report. I wonder if `requests` code I replaced with `aiohttp` had `raise_for_status()` in use and I missed it. Will check.

I think we can also set `raise_for_status=True` on the session or the individual requests as per:
https://docs.aiohttp.org/en/stable/client_reference.html

I'll accept a PR. Lets try setting it to `True` on the aiohttp `session` we create here:
https://github.com/pypa/bandersnatch/blob/master/src/bandersnatch/master.py#L44
- Then get a unittest/integration test to cover this so we don't break it again.

If you don't have time to PR I'll try get to it this week.

windtail:
As my previous hack, if `aiohttp.ClientResponseError`  raised, package will be deleted.

And I just confirmed that `aiohttp.ClientResponseError` was raised by following code

```python
@pytest.mark.asyncio
[...]

PR #496, opened 2020-04-24
Raise error for bad response (#494)

windtail:
I am not sure about the unit test. I think we should only test correct argument is passed to aiohttp, whether aiohttp actually raise the exception should tested in aiohttp itself.


src/bandersnatch/tests/test_master.py
```diff
@@ -69,3 +69,13 @@ async def test_master_doesnt_raise_if_serial_equal(master):
...
+
+
+@pytest.mark.asyncio
+async def test_session_raise_for_status(master):
+    patcher = asynctest.patch("aiohttp.ClientSession", autospec=True)
+    with patcher as create_session:
+        async with master:
+            pass
+        assert len(create_session.call_args_list) == 1
+        assert create_session.call_args_list[0][1]["raise_for_status"]
```
cooperlees, line 81: Cool - this just ensures it's set.
- I was hoping to trigger it somehow on a run to see the delete take place. If you don't have time I'll still accept this.

PR #497, opened 2020-04-26
Update pre commit and other changes

SanketDG:


src/bandersnatch/main.py
```diff
@@ -102,7 +102,7 @@ async def async_main(args: argparse.Namespace, config: ConfigParser) -> int:
         if status_file.exists():
             tmp_status_file = Path(gettempdir()) / "status"
             try:
-                shutil.move(status_file, tmp_status_file)
+                shutil.move(str(status_file), tmp_status_file)
```
cooperlees, line 105: Hmm - I thought this could handle a path-like-object but both 3.6 and 3.7 documentation don't say it can. So thanks!

cooperlees:
P.s. codecov really confuses me. I'm thinking of dropping it :|

src/bandersnatch/main.py
SanketDG: Relevant (it's interesting): https://github.com/python/typeshed/pull/3559

SanketDG:
> Awesome - Hope monkeytype is working. I use to sit 2 rows away from the guy who wrote it at Instagram :)

Oh, that's nice!

> I did plan to fire it up with running tox + the integration test to get some typing for free.

Planning to do soon.

> P.s. codecov really confuses me. I'm thinking of dropping it :|

Yeah, I was also not sure what's going on. But that's fine, every piece of software fails occasionally, some more, some less :D

PR #514, opened 2020-05-12
WIP: [fix] Tune the aiohttp timeout options.

zxkane:
Some TODO items,

- [ ] unit tests
- [ ] make the total timeout configurable


cooperlees:
Did this actually help? I need to reread the definitions but I do think this makes sense. I’m more thinking an hour default tho.

cooperlees:
Found the issue.

cooperlees:
#535 addresses this PR.

zxkane:
@cooperlees , the total timeout only relying on the `socket* timeout` is difficult to tune a reasonable socket timeout. You already saw the downloading case in my network, it cost several hours to complete downloading an artifact sometimes. We will have to increase the socket timeout to `720`,`1080` or even larger to grant enough time downloading the artifacts. But it does not make sense to set the socket timeout to 12 min or 18 min without receiving any data in that long period. It probably impacts the progress of entire mirroring task.

How about adding an additional config option or environment variable for total timeout, which only takes effect if it's given? It would give a chance to make it work in the slow network. Let me know what you think. Thanks.

cooperlees:
As I stated elsewhere, totally for i. As long as the code can fallback to a sane default if it's not in the config file. Please also unittest it.

I had just been awaiting your PR to be finished and wanted something that's better than what we currently have, so I did #535.

cooperlees:
#540 is giving you your config flag.

zxkane:
It really helps. I will try it when new release is published.

Issue #463, opened 2020-03-22
Refactor verify to use package, mirror + master classes


cooperlees:
`verify`, when originally written, used aiohttp as a POC. Now that the core functions use aiohttp and everything is async, we should have verify use the classes and async methods that now exist and reduce duplicated code.

E.g. Package retrieval and syncing binaries + JSON files etc.

PR #541, opened 2020-05-27
Ensure use of `storage_backend` based path wrapper

### LINKED ISSUE #539: url_fetch and verify/delete do not rely on storage backend-based path operations
Operations in the `verify.py` module generally speaking rely on naked calls to `pathlib.Path` on the mirror base directory instead of relying on the storage backend, which could cause issues for non-filesystem based IO

techalchemy:
- When downloading, verifying or deleting content, make sure we
  use the `storage_backend` wrapped path plugin
- This is going to be really annoying and we probably should figure out
  how to centralize these kinds of calls in a more natural way
- Fixes #539

Signed-off-by: Dan Ryan <dan.ryan@canonical.com>

### DIFF
--- a/src/bandersnatch/storage.py
+++ b/src/bandersnatch/storage.py
@@ -60,7 +60,7 @@ class Storage:
             self.configuration = BandersnatchConfig().config
         try:
             storage_backend = self.configuration["mirror"]["storage-backend"]
-        except KeyError:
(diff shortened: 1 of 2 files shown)


cooperlees:
Ahh - I do have #463 open for it. Thought I did.

PR #546, opened 2020-05-31
Enforce mypy on all Core bandersnatch

### LINKED ISSUE #15: PEP484 type annotate all of bandersnatch

cooperlees:
- Set Python 3.6
- Turn on a lot more mypy checks + warnings
- Exclude all plugins for now and focus on main code base

More work to get closer to finally close #15

### DIFF
--- a/src/bandersnatch/configuration.py
+++ b/src/bandersnatch/configuration.py
@@ -48,7 +48,9 @@ class BandersnatchConfig(metaclass=Singleton):
             Path to the configuration file to use
         """
         self.found_deprecations: List[str] = []
-        with importlib.resources.path("bandersnatch", "default.conf") as config_path:
+        with importlib.resources.path(  # type: ignore
+            "bandersnatch", "default.conf"
+        ) as config_path:
             self.default_config_file = str(config_path)
         self.config_file = config_file
         self.load_configuration()
--- a/src/bandersnatch_storage_plugins/filesystem.py
+++ b/src/bandersnatch_storage_plugins/filesystem.py
@@ -97,7 +97,7 @@ class FilesystemStorage(StoragePlugin):
         self.delete_file(filepath_tmp)
 
     @contextlib.contextmanager
-    def update_safe(self, filename: str, **kw: Any) -> Generator[IO, None, None]:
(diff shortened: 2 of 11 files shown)


src/bandersnatch/configuration.py
```diff
@@ -48,7 +48,9 @@ def __init__(self, config_file: Optional[str] = None) -> None:
             Path to the configuration file to use
         """
         self.found_deprecations: List[str] = []
-        with importlib.resources.path("bandersnatch", "default.conf") as config_path:
+        with importlib.resources.path(  # type: ignore
+            "bandersnatch", "default.conf"
+        ) as config_path:
```
cooperlees, line 53: @SanketDG - Any ideas why this error occurs? I haven't dug into it ... Maybe because in 3.6 we use the pulled in dependency?

I might move 4.1 to be >= 3.7.0 since Docker containers make moving to newer python so much easier these days ...

src/bandersnatch_storage_plugins/filesystem.py
```diff
@@ -97,7 +97,7 @@ def rewrite(
         self.delete_file(filepath_tmp)
 
     @contextlib.contextmanager
-    def update_safe(self, filename: str, **kw: Any) -> Generator[IO, None, None]:
+    def update_safe(self, filename: PATH_TYPES, **kw: Any) -> Generator[IO, None, None]:
```
cooperlees, line 100: @techalchemy - This should be legit - my quick looks seems both can be handled or am I missing something?

src/bandersnatch/configuration.py
SanketDG: Yup, the library does not seem to have type hints
https://gitlab.com/python-devs/importlib_metadata/-/issues/10

src/bandersnatch_storage_plugins/filesystem.py
cooperlees: Dan ok'd in IRC. Merging :D

PR #550, opened 2020-06-04
Move http request code for package metadata to master.py

dralley:
It feels like it makes more sense as a utility function there, beside all the other API calls.

### DIFF
--- a/src/bandersnatch/package.py
+++ b/src/bandersnatch/package.py
@@ -9,11 +9,10 @@ from shutil import rmtree
 from typing import TYPE_CHECKING, Dict, List, Optional, Union
 from urllib.parse import unquote, urlparse
 
-from aiohttp import ClientResponseError
 from packaging.utils import canonicalize_name
 
 from . import utils
-from .master import StalePage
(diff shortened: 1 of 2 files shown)


src/bandersnatch/package.py
```diff
@@ -142,17 +141,13 @@ async def sync(self, stop_on_error: bool = False, attempts: int = 3) -> None:
...
-                        if e.status == 404:
-                            logger.info(f"{self.name} no longer exists on PyPI")
-                            return None
-                        raise
+
+                    metadata = await self.mirror.master.get_package_metadata(
+                        self.name, serial=self.serial
+                    )
+                    if not metadata:
+                        return None
```
dralley, line 149: Would a custom error type be better?  e.g. have the ```get_package_metadata``` function ```raise PackageNotFound(f"{self.name} no longer exists on PyPI"``` and then catch and log and return as before?

cooperlees:
Travis reliability has gone downhill lately :|

dralley:
It has, we've had issues with it as well :/

dralley:
Looks like it's good now though

Issue #556, opened 2020-06-06
Fix Documentation to be sphinx 3.0 friendly


cooperlees:
We have 224 warnings and some formatting is off - This was raised by @ichard26.

Also make our CI test better for detecting Documentation breakage!

PR #557, opened 2020-06-06
Fix conf.py so Markdown docs can build properly

ichard26:
Related #556.

## Fix conf.py so Markdown docs can build properly 
Since the bump to Sphinx 3.x, all of the Markdown documentation wasn't building
since the Markdown configuration was outdated. Instead of using source_parsers,
you just add the recommonmark extension. Also for good measure, Sphinx 3.0 < is
now required.

## Add protection for incompatible recommommark versions 
recommonmark version 0.4.0 and below are incompatible with Sphinx 3.x and since
we moved to Sphinx 3.x, a safeguard should be added.

---

Here's the built docs:
https://ichard26-testbandersnatchdocs.readthedocs.io/en/sphinx3/

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -23,6 +23,10 @@ except ImportError:
 # If your documentation needs a minimal Sphinx version, state it here.
 needs_sphinx = "3.0"
 
+# Just a protection against an incompatible version of recommonmark.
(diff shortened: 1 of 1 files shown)


ichard26:
Oops, flake8 doesn't like my code.

PR #561, opened 2020-06-08
Type annotate bandersnatch plugins

ichard26:
Related: #15.

Type annotate the bandersnatch plugins.

@techalchemy, it would be nice if you could review this PR since you know these plugins the best. (Only if you have the time though)

### DIFF
--- a/src/bandersnatch_filter_plugins/prerelease_name.py
+++ b/src/bandersnatch_filter_plugins/prerelease_name.py
@@ -1,5 +1,6 @@
 import logging
 import re
+from typing import Dict
 
 from bandersnatch.filter import FilterReleasePlugin
 
@@ -20,7 +21,7 @@ class PreReleaseFilter(FilterReleasePlugin):
     )
     patterns = None
 
-    def initialize_plugin(self):
+    def initialize_plugin(self) -> None:
         """
         Initialize the plugin reading patterns from the config.
         """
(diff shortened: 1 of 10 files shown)


ichard26:
I also learned `# type: ignore[override]` today. I was a little annoyed with `Signature incompatible with Super type` warnings. Also, trust me, the only reason why it looks *legit* is because of the amount of time I spent on it.

cooperlees:
And I appreciate it. As someone who’s typed a lot of old code, I appreciate the difficulties. Typing code changes how you code moving forward. For the better I believe :)

src/bandersnatch_filter_plugins/prerelease_name.py
```diff
@@ -31,13 +32,13 @@ def initialize_plugin(self):
...
 
-    def filter(self, metadata):
+    def filter(self, metadata: Dict) -> bool:
         """
         Remove all release versions that match any of the specified patterns.
         """
         releases = metadata["releases"]
         for version in list(releases.keys()):
-            if any(pattern.match(version) for pattern in self.patterns):
+            if any(pattern.match(version) for pattern in self.patterns):  # type: ignore
```
cooperlees, line 41: Will be interested to see what the error is here.

ichard26:
```
prerelease_name.py:41: error: Item "None" of "Optional[List[Pattern[str]]]" has no attribute "__iter__" (not iterable)
```

cooperlees: Ahh - makes sense.

Yeah, here we just need to add another check for a None object and handle it. I’ll go back and look at type ignores before closing the issue and see if we can get rid of a few more.

PR #563, opened 2020-06-08
Minor cleanup & factor package retry logic out of Package.sync()

dralley:


dralley:
Still need to fix some of the type annotations

src/bandersnatch/package.py
```diff
@@ -133,68 +141,71 @@ def save_json_metadata(self, package_info: Dict) -> bool:
 
         return True
 
-    async def sync(self, stop_on_error: bool = False, attempts: int = 3) -> None:
-        loop = asyncio.get_event_loop()
-        self.tries = 0
+    async def fetch_metadata(self, attempts: int = 3) -> Any:
```
dralley, line 144: So the reason this type annotation is overly-broad is because ```master.get_package_metadata()``` returns ```Any```, and it does so because ```master.get()``` returns ```Any```, and I'm not sure if we can change the return value there.

But this is also my first experience with mypy and type annotations in Python so let me know if there is anything we can do to get around that.

src/bandersnatch/tests/test_package.py
```diff
@@ -65,19 +66,19 @@ async def test_package_sync_404_json_info_keeps_package_on_non_deleting_mirror(m
 
 
 @pytest.mark.asyncio
-async def test_package_sync_gives_up_after_3_stale_responses(caplog, mirror):
+async def test_package_fetch_metadata_gives_up_after_3_stale_responses(caplog, mirror):
     mirror.master.package_releases = mock.Mock()
     mirror.master.package_releases.return_value = ["0.1"]
     mirror.master.release_urls = mock.Mock()
     mirror.master.release_urls.return_value = []
+    mirror.master.get_package_metadata = mock.AsyncMock(side_effect=StalePage)
```
dralley, line 74: Looks like ```AsyncMock``` doesn't exist in Python 3.6.

cooperlees:
Thanks for this cleanup! I look forward to this being way more usable by pulp and other projects and appreciate your efforts.

PR #586, opened 2020-06-19
Move legacy directory cleanup to mirror.py

dralley:
Since the overall goal is to move operations which touch the filesystem away from the public API, this code should be moved to mirror.py.

I think this makes sense for a lot of other reasons, generally (annotated inline)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -105,3 +105,6 @@ UNKNOWN.egg-info/
 
 # MonkeyType
 monkeytype.sqlite3
+
+# Integration test
+mirrored-files
--- a/src/bandersnatch/mirror.py
+++ b/src/bandersnatch/mirror.py
@@ -5,6 +5,7 @@ import logging
 import os
 import time
 from pathlib import Path
+from shutil import rmtree
 from threading import RLock
 from typing import Awaitable, Dict, List, Optional, Set, Union
 from unittest.mock import Mock
@@ -250,15 +251,16 @@ class Mirror:
 
             await package.sync()
 
+            # Cleanup non normalized name directory
+            await self.cleanup_non_pep_503_paths(package)
+
     async def sync_packages(self) -> None:
         self.package_queue: asyncio.Queue = asyncio.Queue()
         # Sorting the packages alphabetically makes it more predicatable:
         # easier to debug and easier to follow in the logs.
         for name in sorted(self.packages_to_sync):
             serial = self.packages_to_sync[name]
-            await self.package_queue.put(
-                Package(name, serial, self, cleanup=self.cleanup)
-            )
+            await self.package_queue.put(Package(name, serial, self))
 
         sync_coros: List[Awaitable] = [
             self.package_syncer(idx) for idx in range(self.workers)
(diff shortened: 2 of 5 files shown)


src/bandersnatch/package.py
```diff
@@ -46,15 +45,11 @@ def __init__(
         name: str,
         serial: Union[int, str],
         mirror: "Mirror",
-        *,
-        cleanup: bool = False,
```
dralley, line 50: It seemed silly to have two copies of ```cleanup```, where ```mirror.cleanup``` is only used to set ```package.cleanup```

src/bandersnatch/package.py
```diff
@@ -46,15 +45,11 @@ def __init__(
...
     ) -> None:
         self.name = canonicalize_name(name)
         self.raw_name = name
-        self.normalized_name_legacy = utils.bandersnatch_safe_name(name)
```
dralley, line 54: Want to keep legacy things out of the public API if possible, which is easy because this field was *only* used for cleanup

src/bandersnatch/mirror.py
```diff
@@ -290,6 +294,52 @@ def record_finished_package(self, name: str) -> None:
                 ]
                 f.write("\n".join(todo))
 
+    async def cleanup_non_pep_503_paths(self, package: Package) -> None:
+        """
+        Before 4.0 we use to store backwards compatible named dirs for older pip
+        This function checks for them and cleans them up
+        """
+
+        def raw_simple_directory() -> Path:
```
dralley, line 303: These are only used for cleanup so I stuck them inline

src/bandersnatch/tests/test_mirror.py
```diff
@@ -11,9 +12,18 @@
...
+from bandersnatch.package import Package
 from bandersnatch.utils import WINDOWS
 
 
+def touch_files(paths: List[Path]):
+    for path in paths:
+        if not path.parent.exists():
+            path.parent.mkdir(parents=True)
+        with path.open("wb") as pfp:
+            pfp.close()
```
dralley, line 24: This is copied and pasted, which sucks, but it's only temporary until the other parts of ```Package``` are relocated as well.  I can move it somewhere and import from both places if you prefer.

src/bandersnatch/tests/plugins/test_storage_plugins.py
```diff
@@ -594,32 +594,6 @@ def test_plugin_type(self):
         self.assertTrue(isinstance(self.plugin, self.plugin_map[self.backend]))
         self.assertTrue(self.plugin.PATH_BACKEND is self.path_backends[self.backend])
 
-    def test_json_paths(self):
```
dralley, line 597: Dead code, presumably replaced by the code in ```package.py```

src/bandersnatch/storage.py
```diff
@@ -116,16 +112,6 @@ def get_lock(self, path: str) -> filelock.BaseFileLock:
         """
         raise NotImplementedError
 
-    def get_json_paths(self, name: str) -> Sequence[PATH_TYPES]:
```
dralley, line 119: Dead code, presumably replaced by similar code in ```package.py```

.gitignore
```diff
@@ -105,3 +105,6 @@ UNKNOWN.egg-info/
 
 # MonkeyType
 monkeytype.sqlite3
+
+# Integration test
+mirrored-files
```
cooperlees, line 110: I did notice this the other day too. We should get test_runner.py to clean this up with an atexit probably. For another PR.

src/bandersnatch/mirror.py
```diff
@@ -290,6 +292,52 @@ def record_finished_package(self, name: str) -> None:
...
+
+        if not self.cleanup:
+            return
+
+        logger.debug(f"Running Non PEP503 path cleanup for {package.raw_name}")
+        for deprecated_dir in (
+            raw_simple_directory(),
+            normalized_legacy_simple_directory(),
+        ):
+            # Had to compare path strs as Windows did not match path objects ...
```
cooperlees, line 325: Hmm - I feel pathlib should handle that :| ... MIght be a new thing due to the storage abstraction from @techalchemy

src/bandersnatch/mirror.py
```diff
@@ -290,6 +292,52 @@ def record_finished_package(self, name: str) -> None:
...
+                if not deprecated_dir.exists():
+                    logger.debug(f"{deprecated_dir} does not exist. Not cleaning up")
+                    continue
+
+                logger.info(
+                    f"Attempting to cleanup non PEP 503 simple dir: {deprecated_dir}"
+                )
+                try:
+                    rmtree(deprecated_dir)
+                except Exception:
```
cooperlees, line 336: Maybe be a little more specific here - Catch OSError should do.

src/bandersnatch/tests/test_mirror.py
cooperlees: Can’t we just move to utils or a new test_utils.py to save the copy?

src/bandersnatch/tests/test_package.py
```diff
@@ -77,7 +62,7 @@ async def test_package_fetch_metadata_gives_up_after_3_stale_responses(caplog, m
     pkg_name = "foo"
     package = Package(pkg_name, 11, mirror)
 
-    with pytest.raises(Exception):
+    with pytest.raises(StaleMetadata):
```
cooperlees, line 65: Nice. I’m a big fan of more specific exception handling.

src/bandersnatch/mirror.py
cooperlees: FWIW - these use to be used else where but I clean all that up.

Issue #548, opened 2020-06-03
Proposal: A general API for PyPI mirroring tools


dralley:
Proposal
========

Bandersnatch is the top tool for mirroring PyPI. It's well tested in production, PEP-compliant, and generally respectful of PyPI's resources. It would be excellent if Bandersnatch could be refactored to provide a common library for all mirroring tools interacting with PyPI.

Benefits to Bandersnatch:

* More testing
* More eyes on the code
* More stakeholders willing to fix bugs and make improvements in Bandersnatch which will trickle down to their own projects

Benefits to the Python ecosystem:

* Easier to change Warehouse APIs (e.g. remove XML-RPC, add new JSON APIs) if Bandersnatch is the only consumer of those APIs (and other mirroring tools just use Bandersnatch)

What we need
============

I write this as a contributor to one of these other mirroring tools (although we do more "repo management" rather than just mirroring). Here is the functionality we would need from Bandersnatch (including things that might already be do-able):

* The ability to parse Bandersnatch config files and get the configuration data back in a form that we can save as settings in our database, and re-constitute later for use by the mirroring stage
* The ability to filter PyPI packages/releases using the filters Bandersnatch provides
* The following metadata about each package that passes through the filters:
    * filename
    * packagetype
    * name
    * version
    * metadata_version
    * summary
    * description
    * keywords
    * home_page
    * download_url
    * author
    * author_email
    * maintainer
    * maintainer_email
    * license
    * requires_python
    * project_url
    * platform
    * supported_platform
    * requires_dist
    * provides_dist
    * obsoletes_dist
    * requires_external
    * classifiers
* The download url to the actual file in the PyPI datastore

Current Problems
================

Problem: Separation of concerns - the communication with PyPI isn't very separated from internal Bandersnatch-specific behavior.

a) It looks like the primary entry point, Mirror.synchronize(), unavoidably tries to touch the filesystem in various ways that are needed by Bandersnatch but not for someone calling Bandersnatch as a library, such as:

* "todolist"
* "last_modified"
* "statusfile"

b) It doesn't seem that there is a way to ask Bandersnatch to abstain from downloading the package files (or JSON files? unclear on that one) and saving them to disk.

We need to handle downloading the package file ourselves, the only thing we need is the JSON metadata (as a dict or string) and the package download URL.

[deleted tangent about *why* we need it]

Solution
========

Separate the concerns a little bit better. I think the following suggestions would be the *minimum* necessary changes:

[...]

PR #591, opened 2020-06-25
Generic API for third-party mirroring tools

dralley:
Attempting to implement a generic API for third-party tools that want to mirror or collect information from Warehouse/PyPI.

https://github.com/pypa/bandersnatch/issues/548


src/bandersnatch/mirror.py
```diff
@@ -249,18 +251,73 @@ async def package_syncer(self, idx: int) -> None:
                 logger.debug(f"Package syncer {idx} emptied queue")
                 break
 
-            await package.sync()
+            try:
+                metadata = await package.fetch_metadata(attempts=3)
+                # Don't save anything if our metadata filters all fail.
```
dralley, line 256: All this I want to eventually move back into a function on this side (as opposed to on the package)

src/bandersnatch/mirror.py
```diff
@@ -494,6 +551,11 @@ def _save(self) -> None:
         self.statusfile.write_text(str(self.synced_serial), encoding="ascii")
 
 
+# class MetadataWriter:
```
dralley, line 554: Meant to delete this, whoops.

src/bandersnatch/mirror.py
dralley: In the process of refactoring though ```Mirror``` might become a bit of a god-class.  Hopefully temporarily.

dralley:
Everything that saves metadata and touches the filesystem I basically want to move into one class.  Right now that class is ```Mirror```.

My rough idea of a long-term plan is to have one class responsible for fetching and filtering metadata, which exposes a queue of packages to be consumed by a second class that is responsible for saving metadata.

Or in the case of Pulp, we would just consume that queue directly.

In the process of refactoring though Mirror might temporarily become a bit of a god-class.

src/bandersnatch/package.py
```diff
@@ -40,50 +28,30 @@ def __str__(self) -> str:
...
-    @property
-    def json_file(self) -> Path:
-        return Path(self.mirror.webdir / "json" / self.name)
+        self._metadata = None
 
     @property
-    def json_pypi_symlink(self) -> Path:
-        return Path(self.mirror.webdir / "pypi" / self.name / "json")
+    def metadata(self) -> dict:
+        assert self._metadata is not None, "Must fetch metadata before accessing it"
```
cooperlees, line 41: asserts in optimized python won't be evaluated at runtime ...

src/bandersnatch/package.py
```diff
@@ -40,50 +28,30 @@ def __str__(self) -> str:
...
+        return self._metadata
 
     @property
-    def simple_directory(self) -> Path:
-        if self.mirror.hash_index:
-            return Path(self.mirror.webdir / "simple" / self.name[0] / self.name)
-        return Path(self.mirror.webdir / "simple" / self.name)
+    def info(self) -> dict:
+        return self.metadata["info"]
```
cooperlees, line 46:
```suggestion
        return self._metadata["info"]
```

src/bandersnatch/package.py
```diff
@@ -40,50 +28,30 @@ def __str__(self) -> str:
...
-        symlink_dir = self.json_pypi_symlink.parent
-        symlink_dir.mkdir(exist_ok=True)
-        # Lets always ensure symlink is pointing to correct self.json_file
-        # In 4.0 we move to normalized name only so want to overwrite older symlinks
-        if self.json_pypi_symlink.exists():
-            self.json_pypi_symlink.unlink()
-        self.json_pypi_symlink.symlink_to(self.json_file)
+    @property
+    def last_serial(self) -> int:
+        return self.metadata["last_serial"]
```
cooperlees, line 50:
```suggestion
        return self._metadata["last_serial"]
```

src/bandersnatch/mirror.py
```diff
@@ -493,6 +587,174 @@ def _load(self) -> None:
     def _save(self) -> None:
         self.statusfile.write_text(str(self.synced_serial), encoding="ascii")
 
+    def gen_data_requires_python(self, release: Dict) -> str:
+        if "requires_python" in release and release["requires_python"] is not None:
+            return f' data-requires-python="{html.escape(release["requires_python"])}"'
+        return ""
+
+    def generate_simple_page_for_package(self, package: Package) -> str:
```
cooperlees, line 595: I don't get the advantage here? Can't you just not call this when using a Package() object?

src/bandersnatch/package.py
dralley: What do you mean? I don't think Python ever skips asserts (unlike languages like C, Rust)

src/bandersnatch/mirror.py
dralley: My thought process is that ```Package``` is/should be just a data structure with metadata about the package and releases, while the task of implementing the simple API (including the index) should be handled by a single dedicated class.

So, right now, the task of writing HTML is spit between package.py and mirror.py - mirror.py writes the [index page](https://github.com/pypa/bandersnatch/blob/master/src/bandersnatch/mirror.py#L373). My thought is that there should just be a ```SimpleMetadataWriter``` class or something which is responsible for both - it takes a list/queue/iterable of packages and turns it into a full repository, handling all the HTML writing both for the packages and for the index.

I'm definitely not married to any one particular idea though, there's a reason I ask before putting a ton of time into going in a particular direction :)

src/bandersnatch/package.py
dralley: So I've just learned about python's ```-O``` option, but I don't think it's commonly used because it does pretty much nothing.

src/bandersnatch/package.py
```diff
@@ -1,91 +1,56 @@
...
         self.raw_name = name
         self.serial = serial
-        self.mirror = mirror
+
+        self._metadata = None
 
     @property
-    def json_file(self) -> Path:
-        return Path(self.mirror.webdir / "json" / self.name)
+    def metadata(self) -> dict:
```
cooperlees, line 28:
```suggestion
    def metadata(self) -> Dict:
```

NIT: Don't type with standard typing objects - Using Generics types is preferred from the typing module: https://docs.python.org/3/library/typing.html#generics

src/bandersnatch/package.py
cooperlees: Not a fan, but can live with it. I feel we should raise a `NeedMetadataError` from your new errors.py eventually.

src/bandersnatch/package.py
```diff
@@ -1,91 +1,56 @@
...
+        assert self._metadata is not None, "Must fetch metadata before accessing it"
+        return self._metadata
 
     @property
-    def json_pypi_symlink(self) -> Path:
-        return Path(self.mirror.webdir / "pypi" / self.name / "json")
+    def info(self) -> dict:
```
cooperlees, line 33:
```suggestion
    def info(self) -> Dict:
```

src/bandersnatch/tests/test_mirror.py
```diff
@@ -492,7 +492,7 @@ def test_validate_todo(mirror: Mirror) -> None:
 @pytest.mark.asyncio  # type: ignore
 async def test_cleanup_non_pep_503_paths(mirror: Mirror) -> None:
     raw_package_name = "CatDogPython69"
-    package = Package(raw_package_name, 11, mirror)
+    package = Package(raw_package_name, mirror)
```
cooperlees, line 495: Is this just to test serial = 0 default now?

src/bandersnatch/mirror.py
```diff
@@ -596,36 +575,36 @@ def _save(self) -> None:
...
+    def json_pypi_symlink(self, package_name: str) -> Path:
         return Path(self.webdir / "pypi" / package_name / "json")
 
     def simple_directory(self, package: Package) -> Path:
         if self.hash_index:
             return Path(self.webdir / "simple" / package.name[0] / package.name)
         return Path(self.webdir / "simple" / package.name)
 
-    def save_json_metadata(self, package_info: Dict) -> bool:
+    def save_json_metadata(self, package_info: Dict, name: str) -> bool:
```
dralley, line 586: Probably what we should do here is just take a package as a parameter, and extract both from that.

src/bandersnatch/tests/test_mirror.py
dralley: Flipped that unintentionally, I fixed it

src/bandersnatch/mirror.py
```diff
@@ -193,6 +115,144 @@ def _filter_packages(self) -> None:
...
+                    "Cancelling, all downloads are forcibly stopped, data may be "
+                    + "corrupted. Serial will not be saved to disk. "
+                    + "Next sync will start from previous serial"
+                )
+        except (ValueError, TypeError):
+            # This is for when self.packages_to_sync isn't of type Dict[str, int]
+            self.errors = True
+
+    def finalize_sync(self) -> None:
+        return None
```
dralley, line 178: I did this initially but actually I think we should force the user to implement it, even if it's just a no-op.  That way it'll be harder to accidentally misuse.

dralley:
The API around the ```determine_packages_to_sync()``` method is still kind of messy, and I think a lot of it will end up being copied and pasted into various implementations.  It would be nice if we can find a way to avoid that.  Doesn't necessarily need ot happen in this PR, though, it's already substantial enough.

dralley:
@gerrod3 Could you squash the "temp" commit into the "BandersnatchMirror now handles all file I/O" commit when you get a chance?

src/bandersnatch/mirror.py
```diff
@@ -193,6 +109,139 @@ def _filter_packages(self) -> None:
...
+        raise NotImplementedError()
+
+    def on_error(self, exception: BaseException, **kwargs: Dict) -> None:
+        raise NotImplementedError()
+
+
+class BandersnatchMirror(Mirror):
+
+    need_index_sync = True
+    json_save = False  # Whether or not to mirror PyPI JSON metadata to disk
```
dralley, line 167: A lot of these class variables are overwritten by ```__init__```.   Let's double check the defaults below to make sure they match up, move the comments, and then delete the superfluous definitions here.

src/bandersnatch/mirror.py
```diff
@@ -241,42 +291,73 @@ async def determine_packages_to_sync(self) -> None:
...
+        if isinstance(exception, KeyboardInterrupt):
             # Setting self.errors to True to ensure we don't save Serial
             # and thus save to disk that we've had a successful sync
-            self.errors = True
             logger.info(
                 "Cancelling, all downloads are forcibly stopped, data may be "
                 + "corrupted. Serial will not be saved to disk. "
                 + "Next sync will start from previous serial"
             )
+        elif isinstance(exception, TypeError) or isinstance(exception, ValueError):
```
dralley, line 334: Since we've separated this from it's context, it would probably be a good idea to have a comment here describing where these errors are coming from.

src/bandersnatch/mirror.py
```diff
@@ -193,6 +109,139 @@ def _filter_packages(self) -> None:
...
+
+            sync_coros: List[Awaitable] = [
+                self.package_syncer(idx) for idx in range(self.workers)
+            ]
+            try:
+                await asyncio.gather(*sync_coros)
+            except KeyboardInterrupt as e:
+                self.on_error(e)
+        except (ValueError, TypeError) as e:
+            # This is for when self.packages_to_sync isn't of type Dict[str, int]
```
dralley, line 154: Does this happen as part of the normal control flow?  Under what circumstances?  If this happens semi-regularly, then making a distinguished exception type would be preferable I think.

But I'm missing some context here

gerrod3: Actually this error check is for some poorly written tests.  No where in the normal control flow should this error occur. It happens when one manually sets the self.packages_to_sync to any value that doesn't follow the type Dict[str, Union[int, str]], which some of the tests do purposely set to incorrect values.

dralley: Oof.  Let's definitely call that out as a TODO then.

As an aside, that doesn't really make sense, the entire point of adding type annotations is to catch bugs like that so that unit tests for that sort of thing are no longer necessary.  So hopefully it can be ditched.

PR #592, opened 2020-06-26
type annotate tests

rkm:
hi!

This PR contains my work towards closing #15 . All the main tests now pass mypy, except for the plugin tests which I will continue to work on. As discussed in #15, there are some `# type: ignore`s that need to be worked through at some point once pytest supports typing better.


src/bandersnatch/tests/test_main.py
```diff
@@ -54,23 +59,23 @@ def test_main_cant_create_config(caplog, tmpdir):
...
-    config_path = tmpdir.join("unittest.conf")
+    config_path = tmpdir / "unittest.conf"
     config_path.write_text("\n".join(config_lines), encoding="utf-8")
     sys.argv = ["bandersnatch", "-c", str(config_path), "mirror"]
     assert config_path.exists()
     main(asyncio.new_event_loop())
     (homedir, master), kwargs = mirror_mock.call_args_list[0]
 
-    assert "/srv/pypi" == homedir
+    assert Path("/srv/pypi") == homedir
```
cooperlees, line 78: Thanks for this change on Mirror - just hope non fikeaystem objects can handle it - cc: @techalchemy

src/bandersnatch/tests/test_utils.py
```diff
@@ -60,16 +62,16 @@ def test_find_files():
...
         assert found_files == expected_found_files
 
 
-def test_rewrite(tmpdir, monkeypatch):
+def test_rewrite(tmpdir: Path, monkeypatch: MonkeyPatch) -> None:
     monkeypatch.chdir(tmpdir)
     with open("sample", "w") as f:
         f.write("bsdf")
-    with rewrite("sample") as f:
+    with rewrite("sample") as f:  # type: ignore
```
cooperlees, line 74: What is the type error here ... weird. I’ll look after merge.

rkm: The error is:
```py
src/bandersnatch/tests/test_utils.py: note: In function "test_rewrite":
src/bandersnatch/tests/test_utils.py:74: error: Incompatible types in
assignment (expression has type "IO[Any]", variable has type "TextIO")
        with rewrite("sample") as f:
```

ichard26: The problem is that in the open() call above has the return type `typing.TextIO` because mypy sees the `"w"` and returns a more specific type. The rewrite function can take any mode and then pass it along to `tempfile.NamedTemporaryFile`. Since the mode is defined by a variable, mypy just sets the return type of `tempfile.NamedTemporaryFile` and subsequently `rewrite` to `typing.IO[Any]`.¹

A solution that comes to mind is just adding a `f: IO[Any]` type annotation before the first with statement so `f` isn't inferred as a subclass (`typing.TextIO`) of `typing.IO`.

---
¹. Quoting from Guido van Rossum https://github.com/python/typeshed/issues/1229#issuecomment-304932714

cooperlees: Can't trust that guy, but that's what I would have tried too. Thanks @ichard26

PR #593, opened 2020-06-27
type annotate plugin tests

### LINKED ISSUE #15: PEP484 type annotate all of bandersnatch

rkm:
this finishes-off the test annotations and closes #15! I also refactored-out the `mock_config` function that was duplicated in a few places -- hope that's ok.

I was having trouble running pre-commit through tox locally, so ended up missing black & flake8 in my commits until the end -- you may want to squash-merge.


cooperlees:
Would love to know why tox was unable to run pre-commit or work - What was the error? can you open an issue?

rkm:
Thanks for the merge! Hope I can help out more in future.

> Would love to know why tox was unable to run pre-commit or work - What was the error? can you open an issue?

It was due to something broken in my local environment which I wasn't able to track down. I blew away my local pip and pre-commit caches and it worked again though. Not sure there's enough info there for a bug report unfortunately!

PR #599, opened 2020-07-02
Refactored filter loading into seperate class

gerrod3:
Moved the filter loading into a class that loads all available plugins at once to avoid reading the config file everytime a plugin group wants to be loaded.

### DIFF
--- a/src/bandersnatch/mirror.py
+++ b/src/bandersnatch/mirror.py
@@ -15,7 +15,7 @@ from packaging.utils import canonicalize_name
 
 from . import utils
 from .configuration import validate_config_values
-from .filter import filter_project_plugins, filter_release_plugins
+from .filter import LoadedFilters
 from .master import Master
 from .package import Package
 from .storage import storage_backend_plugins
@@ -82,6 +82,7 @@ class Mirror:
         self.homedir = self.storage_backend.PATH_BACKEND(homedir)
         self.lockfile_path = self.homedir / ".lock"
         self.master = master
+        self.filters = LoadedFilters(load_all=True)
         self.stop_on_error = stop_on_error
         self.json_save = json_save
         self.hash_index = hash_index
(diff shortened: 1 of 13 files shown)


src/bandersnatch/mirror.py
```diff
@@ -82,6 +82,7 @@ def __init__(
         self.homedir = self.storage_backend.PATH_BACKEND(homedir)
         self.lockfile_path = self.homedir / ".lock"
         self.master = master
+        self.filters = LoadedFilters(True)
```
cooperlees, line 85:
```suggestion
        self.filters = LoadedFilters(load_all=True)
```
Maybe just so people know what the bool does without having to go back to filter.py like I did

I've never understood this whole plugin complexity, if we always load all the modules, why do we both having all this extra code? Or am I missing something here?

src/bandersnatch/package.py
```diff
@@ -115,13 +110,15 @@ async def fetch_metadata(self, attempts: int = 3) -> Any:
                 )
                 raise StaleMetadata(package_name=self.name, attempts=attempts)
 
-    async def sync(self, attempts: int = 3) -> None:
+    async def sync(self, filters: "LoadedFilters", attempts: int = 3) -> None:
```
cooperlees, line 113: If we have other external people using this API, this is a pretty big change.

src/bandersnatch/mirror.py
gerrod3: Well currently the plugin groups are loaded when they are first about to be used and then stored in a global variable. However, Mirror loads all of the filter groups anyway so I think it makes sense to just load all of them for the mirror class right away, but having the option to load only the groups they need is useful I think for others using the API.

src/bandersnatch/package.py
gerrod3: I agree.  I'm trying to coordinate my changes with Daniel's PR and since we want to remove the Mirror reference from Package but still allow Package to do its own filtering this is the solution we settled on.  Any ideas for how to improve this would be helpful.

PR #608, opened 2020-07-08
Refactored the removal of releases for release_plugins to happen inside of Package

gerrod3:
Plugins apart of the release_plugin group no longer remove the releases that are passed into the filter.  Release removal is done now in a new function of the Package class ._filter_all_releases().  Also changed the plugin group of the exclude_platform filter to release_file_plugin.

Tests updated to account for new API.  Corrected a linux platform specifier in TestExcludePlatformFilter based on https://github.com/pypa/warehouse/pull/2010

### DIFF
--- a/src/bandersnatch_filter_plugins/latest_name.py
+++ b/src/bandersnatch_filter_plugins/latest_name.py
@@ -1,8 +1,8 @@
 import logging
 from operator import itemgetter
-from typing import Dict, List, Optional, Set, Union
+from typing import Dict, Sequence, Tuple, Union
 
-from packaging.version import parse
+from packaging.version import LegacyVersion, Version, parse
 
 from bandersnatch.filter import FilterReleasePlugin
 
@@ -16,6 +16,7 @@ class LatestReleaseFilter(FilterReleasePlugin):
 
     name = "latest_release"
     keep = 0  # by default, keep 'em all
+    latest: Sequence[str] = []
 
     def initialize_plugin(self) -> None:
         """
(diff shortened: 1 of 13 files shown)


src/bandersnatch_filter_plugins/latest_name.py
```diff
@@ -33,44 +34,31 @@ def initialize_plugin(self) -> None:
...
-            # never remove the stable/official version
-            latest[0] = current_version
+            if before <= self.keep:
+                # not enough releases: do nothing
+                return True
 
-        logger.debug(f"old {versions}")
-        logger.debug(f"new {latest}")
+            versions_pair = map(lambda v: (parse(v), v), versions)
+            self.latest = sorted(versions_pair)[-self.keep :]  # noqa: E203
```
cooperlees, line 55: We should add E203 to our flake8 config for black. (I thought we did)

src/bandersnatch_filter_plugins/latest_name.py
```diff
@@ -16,6 +16,7 @@ class LatestReleaseFilter(FilterReleasePlugin):
 
     name = "latest_release"
     keep = 0  # by default, keep 'em all
+    latest: Sequence = []
```
cooperlees, line 19: Do we know what's always going to be in the sequence? If so lets add it - e.g. Sequence[str]

src/bandersnatch_filter_plugins/latest_name.py
```diff
@@ -33,44 +34,31 @@ def initialize_plugin(self) -> None:
         if self.keep > 0:
             logger.info(f"Initialized latest releases plugin with keep={self.keep}")
 
-    def filter(self, metadata: Dict) -> Optional[bool]:  # type: ignore[override]
```
cooperlees, line 36: Love getting rid of the optional here!

PR #616, opened 2020-07-13
Change to re.match usage to re.search

lps-rocks:
Using re.search instead of re.match allows for non regex formatted matching (static strings) and non-anchored regex matching. This also eliminates the need for .* all over the place when making patterns.

### DIFF
--- a/src/bandersnatch_filter_plugins/metadata_filter.py
+++ b/src/bandersnatch_filter_plugins/metadata_filter.py
@@ -118,7 +118,7 @@ class RegexFilter(Filter):
                 results.append(True)
                 continue
             for value in values:
-                results.append(pattern.match(value))
+                results.append(pattern.search(value))
         return any(results)
 
     def _match_all_patterns(
@@ -129,7 +129,7 @@ class RegexFilter(Filter):
             if nulls_match and not values:
                 results.append(True)
                 continue
-            results.append(any(pattern.match(v) for v in values))
+            results.append(any(pattern.search(v) for v in values))
         return all(results)
 
     def _match_none_patterns(


dralley:
I'm not entirely sure how I feel about this either, the explicitness of the current approach makes a lot of sense to me.

Regardless of whether it's accepted or not, I feel like this probably shouldn't be merged until the next major release of Bandersnatch, since it's a reasonably significant change in behavior.

pombredanne:
@lps-rocks re.search is great but it may actually find things that were not at all wanted. IMHO a filter cannot really accommodate the search semantics as it may cast implicitly too wide a net at times... and there is no way to really when that happened.

That said, a few tests would be great to check what it would mean.

pombredanne:
@lps-rocks also in all cases because of that a change in API, it might be better as a separate option, possibly introducing a `search` "match_patterns" value https://github.com/pypa/bandersnatch/blob/0e6c78aca969d5e0de105a9748584d7b6c777746/src/bandersnatch_filter_plugins/metadata_filter.py#L74

This way this is feature is additive and elective and is not changing the API for existing users.

cooperlees:
O yeah, totally agree with the others. If you add config to toggle from the default of match to search I'll accept this. Bandersnatch is used in production type environments and I don't want to pop unexpected changes on these people.

Thanks for the idea tho and I get your goals. So please rebase and add the config boolean and I think that's acceptable. Please default to current behavior.

lps-rocks:
Yeah this is definitely a breaking change IMHO so waiting till the next major version and noting it in the upgrade notes is a good idea. 

I think adding a config option to toggle which matching method gets used is probably the best idea.

cooperlees:
With the config toggle, I see no reason to wait to a major release. I'll push with 4.1 that I need to pull my finger out and release.

cooperlees:
Is there interest to do this? Or should we close this PR?

lps-rocks:
It's on my list. Just, low on the list.

cooperlees:
That's fine, just wanted to insure there was at least intent. Thanks.

cooperlees:
Please feel free to reopen if you decide to work on this PR. Thanks!

PR #626, opened 2020-07-28
Added new allowlist filter

gerrod3:
The user can add project specifiers to whitelist packages if they enable this new plugin, 'allowlist_release'.


gerrod3:
> If possible can you add some examples to https://github.com/pypa/bandersnatch/blob/master/docs/filtering_configuration.md please :D

I added some more text to show how to use the new filter. Tell me what you think @cooperlees

PR #628, opened 2020-07-29
The PyPA has adopted the PSF code of conduct

hugovk:
For details, see:

* https://discuss.python.org/t/implementing-pep-609-pypa-governance/4745


hugovk:
Ah, good idea and good catch! Thanks, updated!

PR #629, opened 2020-07-29
Casefold package names in blacklist/whitelist plugins

### LINKED ISSUE #284: Whitelist Plugin Not Mirroring All Listed Packages
Whitelist Plugin misses several packages during sync.

I am using --force-check as I am trying to make an agile package request system. A script runs to add dependencies of requested packages to the whitelist in the configuration file.

There are currently 28 packages listed in the whitelist and only 22 installed. 
Example Packages Not being mirrored: flask, async-timeout

I am mirroring all architectures and all versions. This is not a dependency/versioning issue as the problem is directly correlated to the whitelist plugin not pulling every package listed.

Let me know if there is any information you might need to have for troubleshooting help.

lepaperwan:
This PR aims to fix #284 by casefolding package names listed in the bandersnatch configuration as well as those obtained from the XML-RPC when checking for a match.

Also fixes copy-pasta remnants in the whitelist plugin comments/docstrings.

### DIFF
--- a/requirements_test.txt
+++ b/requirements_test.txt
@@ -7,7 +7,7 @@ flake8==3.8.3
 flake8-bugbear==20.1.4
 freezegun==0.3.15
 pre-commit==2.6.0
-pytest==6.0.0rc1
+pytest==6.0.0
 pytest-asyncio==0.14.0
 pytest-timeout==1.4.2
 setuptools==49.2.0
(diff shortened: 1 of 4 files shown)


dralley:
@lepaperwan Let us know if you need help with the unit test

lepaperwan:
This fixes the naive casefolding to effectively implement @cooperlees suggestion.

Also adds a unittest case to ensure non-regression. I've tried to avoid mocking and instead rely on the `Filter` interface, if something needs changing feel free to let me know :)

cooperlees:
It seems XMLRPC is very unhappy today. I've reported it to warehosue peoples, but I need to debug more.

PR looks good. I'll go through it tonight.

PR #632, opened 2020-07-30
More inclusive naming

### LINKED ISSUE #569: More inclusive naming for whitelist and blacklist plugins
Can we use other terms than `whitelist` and `blacklist` to represent the list of packages we are allowing and blocking? I know these are only names and that these terms have established meanings within tech, but I feel that the association of good with white and bad with black has racist connotations and that we could use equally informative alternatives - e.g. `allow` and `deny` (following the convention with firewall rules) or `approved` and `forbidden`.

We're looking to make this change for all our internal references to whitelists and blacklists in one of our projects that uses bandersnatch to enforce these for Python packages, and we are happy to do the work to make equivalent changes to bandersnatch if they are agreed by the community. We recognise that we are new to the bandersnatch community and are looking to start a conversation on this and achieve a community consensus.

I recognise that we'll need to manage the introduction of any agreed change in terminology carefully to not break people's existing installations and will of course follow the standard bandersnatch practices for this.

rkm:
Hi, thought I'd have a go at fixing #569!

This PR replaces all internal usage of whitelist/blacklist with allowlist/denylist. The only remaining references are:

-    in `docs/filtering_configuration.md`. For now, I've just added a deprecation warning to the top of the page
-    in `README.md`. Do we want to add a notice here as well?
-    in `setup.cfg`. I believe this is the only public usage of the old terms(?), so once 5.0 is published these can simply be updated. (The new allowlist_release plugin added in #626 is already named properly).

Some of the docs now read a bit strange, i.e. "to blacklist" sounds ok, but the replacement of "to denylist" looks odd. Should we reword these bits as well?

Should I add a test for the deprecation warning as well?

Fixes #569


rkm:
I think I also need to take into account existing configs -- will investigate further.

rkm:
Apologies, should have marked this as a draft... should be ready to go now.

I think I found (and fixed) a bug as well: https://github.com/pypa/bandersnatch/pull/632/commits/5154126d711d533efa38b7b02b9dd6a485b444fe. The project filters were storing the name+version instead of just the name.

src/bandersnatch/filter.py
```diff
@@ -29,6 +33,7 @@ class Filter:
     """
 
     name = "filter"
+    deprecated_name: Optional[str] = None
```
cooperlees, line 36:
```suggestion
    deprecated_name: str = “”
```
NIT Pick: I see no benefit of non versus a no truths (false) empty str
- won’t block on it tho

src/bandersnatch/tests/plugins/test_allowlist_name.py
```diff
@@ -107,6 +107,28 @@ def test__filter__nomatch_package(self) -> None:
...
+packages =
+    foo==1.2.3
+"""
+        )
+
+        mirror = Mirror(Path("."), Master(url="https://foo.bar.com"))
+        mirror.packages_to_sync = {"foo": "4.5.6"}
+        mirror._filter_packages()
+
+        self.assertIn("foo", mirror.packages_to_sync.keys())
```
cooperlees, line 130: Am I misunderstanding, The allowlist only allows version 1.2.3 shouldn’t foo not exist in the keys() generator/list ?

src/bandersnatch/tests/plugins/test_denylist_name.py
```diff
@@ -104,8 +104,30 @@ def test__filter__nomatch_package(self) -> None:
...
+packages =
+    foo==1.2.3
+"""
+        )
+
+        mirror = Mirror(Path("."), Master(url="https://foo.bar.com"))
+        mirror.packages_to_sync = {"foo": "4.5.6"}
+        mirror._filter_packages()
+
+        self.assertIn("foo", mirror.packages_to_sync.keys())
```
cooperlees, line 127: I agree this should exist due to specifying 1.2.3

src/bandersnatch/tests/test_configuration.py
```diff
@@ -142,6 +143,31 @@ def test_validate_config_values(self) -> None:
...
+        # Test warning if old plugins used
+        instance.SHOWN_DEPRECATIONS = False
+        with open(config_file, "w") as f:
+            f.write("[whitelist]\npackages=foo\n")
+        instance.load_configuration()
+        with warnings.catch_warnings(record=True) as w:
+            instance.check_for_deprecations()
+            instance.check_for_deprecations()
+            # Assert we only throw 1 warning
+            self.assertEqual(len(w), 1)
```
cooperlees, line 169: Great test! Thanks!

tox.ini
```diff
@@ -4,7 +4,7 @@ envlist = lint,py3
 [testenv]
 passenv = CI TRAVIS TRAVIS_*
 commands =
-    coverage run -m pytest
+    coverage run -m pytest {posargs}
```
cooperlees, line 7: Just so I understand - What’s this allowed you to achieve?

dralley:
>Some of the docs now read a bit strange, i.e. "to blacklist" sounds ok, but the replacement of "to denylist" looks odd. Should we reword these bits as well?

Some of the other [suggested](https://github.com/pypa/bandersnatch/issues/569#issuecomment-646182170) terms such as "blocklist" sound marginally better, although ultimately it doesn't matter a ton.  "blocklist" does seem to be gaining traction as a replacement term though.

cooperlees:
Also, sorry, but the case folding PR has created merge conflicts. Can you fix those and I'll merge faster to avoid ... sorry!

if you don't have time say so and I'll add the commit if need be.

rkm:
Hi, thanks for the review! I have some time today so should be able to fix-up the conflicts.

Also now thinking that "blocklist" is a more sensible choice over "denylist", so I'll make that change unless there are any objections?

cooperlees:
I also like blocklist. Thanks!

tox.ini
rkm: This allows us to pass positional arguments along to pytest through tox:
```bash
$ tox -e py3 -- -k foo
```
Unless I'm missing something this isn't currently possible?

src/bandersnatch/tests/plugins/test_allowlist_name.py
rkm: This is to test the bug I believe currently exists (and my fix for it). `AllowlistProject` sates that it only cares about package names and not versions ("This plugin only processes packages, if the line in the packages configuration contains a PEP440 specifier it will be processed by the allowlist release filter").  Currently though it stores the whole filter line including any version specifier, and so will filter any packages not matching that exact name+version.

The corresponding code to this test is [here](https://github.com/pypa/bandersnatch/pull/632/files#diff-9b88701e7c64d1876776b5fb0b0624d5L61-R63)

src/bandersnatch/filter.py
rkm: No worries, have updated this.

src/bandersnatch/tests/plugins/test_allowlist_name.py
rkm: I'm maybe confusing myself over what the intended behaviour is here, but with only the `allowlist_project` plugin enabled, only the package names should be checked? Enabling the `allowlist_release` plugin will then also check the versions? The docs don't specifically mention what the difference is.

gerrod3: Yes @rkm, you are correct, allowlist_project filters only on package names and allowlist_release enables filtering for versions.  The filtering occurs in two stages, before the metadata is downloaded and after the metadata is downloaded.  Project filters (allowlist_project and blocklist_project) occur before the metadata of the package's are downloaded and thus reduce the amount of data that needs to be synced, but they can only filter off of the package's name since you need the metadata to filter off of release info like the version specifiers.  Technically allowlist_release and blocklist_release can perform the same job of their project variants, but you lose some efficiency if you don't care about the versions.

Issue #636, opened 2020-08-02
JSON symlinks to webdir/json/PKG aren't relative


indrat:
The symlinks created in mirror.webdir / pypi/ PKG / json to mirror.webdir / web / json / PKG are not relative making moving the location of the mirror problematic.

cooperlees:
Merged fix.

PR #637, opened 2020-08-02
Make webdir/pypi/json/PKG symlinks relative.

indrat:
In order to assist changing the mirror location or maintaining an
offline mirror at a different path symlinks to webdir/json/PKG should be
relative, addresses #636

### DIFF
--- a/src/bandersnatch/package.py
+++ b/src/bandersnatch/package.py
@@ -2,6 +2,7 @@ import asyncio
 import hashlib
 import html
 import logging
+import os
 import sys
 from json import dump
 from pathlib import Path
@@ -103,7 +104,9 @@ class Package:
         # In 4.0 we move to normalized name only so want to overwrite older symlinks
         if self.json_pypi_symlink.exists():
             self.json_pypi_symlink.unlink()
-        self.json_pypi_symlink.symlink_to(self.json_file)
+        self.json_pypi_symlink.symlink_to(
+            os.path.relpath(self.json_file, self.json_pypi_symlink.parent)
+        )
 
         return True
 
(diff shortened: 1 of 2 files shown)


src/bandersnatch/package.py
```diff
@@ -103,7 +104,9 @@ def save_json_metadata(self, package_info: Dict) -> bool:
         # In 4.0 we move to normalized name only so want to overwrite older symlinks
         if self.json_pypi_symlink.exists():
             self.json_pypi_symlink.unlink()
-        self.json_pypi_symlink.symlink_to(self.json_file)
+        self.json_pypi_symlink.symlink_to(
+            os.path.relpath(self.json_file, self.json_pypi_symlink.parent)
```
cooperlees, line 108: I feel we could use the pathlib method and avoid importing os 
```suggestion
            self.json_file.relative_to(self.json_pypi_symlink.parent)
```

indrat: Unfortunately pathlib.Path.relative_to() won't compute a relative path using a common prefix it only works if the argument to relative_to is a parent of the path, otherwise it raises a ValueError:
```
  File "relative_to.py", line 22, in <module>
    dest.relative_to(src)
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/pathlib.py", line 899, in relative_to
    raise ValueError("{!r} does not start with {!r}"
ValueError: '/var/www/json/PKG' does not start with '/var/www/pypi/PKG/json'
```

cooperlees: You're right - Disappointed that logic does not exist in pathlib - I might open a bug and add it, maybe with a slightly different API.

POC making it work: https://pastebin.com/zTVKFBam
https://pastebin.com/zTVKFBam

PR #638, opened 2020-08-02
Fix passing package info to filters in verify action.

indrat:
Plugin filters appear to now require the full pkg dict rather than just the "info" section.


indrat:
I'm afraid I dont have the time.

PR #641, opened 2020-08-06
Fixed allow/blocklist_release filtering prereleases

gerrod3:
allowlist_release and blocklist_release were using the default construction of the Requirements object which creates a SpecifierSet in order to check requirements.  This SpecifierSet has a field 'prereleases' that gets set to false by default [0] and thus makes any comparisons against prerelease version specifiers return false even if the version would normally fall within the SpecifierSet.  Because of this behaviour allowlist_release and blocklist_release were filtering prereleases even if the prerelease filter wasn't set.

[0] https://packaging.pypa.io/en/latest/specifiers/#packaging.specifiers.SpecifierSet


cooperlees:
CI is all broken due to XMLRPC issues ...

PR #651, opened 2020-08-11
Fixed name parsing issue for allow/blocklist project filters

gerrod3:
Allowlist/blocklist project filters now create the Requirement object first before canonicalizing the package's name. This prevents canonicalize_name() from messing with the package's version specifier, if it was present, which could have caused a parse error inside the Requirement object if the package had the specifier "~=". Also, allowlist/blocklist release filters now how their package names canonicalized before checking to prevent casing mismatches.

### DIFF
--- a/src/bandersnatch/tests/plugins/test_blocklist_name.py
+++ b/src/bandersnatch/tests/plugins/test_blocklist_name.py
@@ -121,10 +121,39 @@ packages =
         )
 
         mirror = Mirror(Path("."), Master(url="https://foo.bar.com"))
-        mirror.packages_to_sync = {"foo": "4.5.6"}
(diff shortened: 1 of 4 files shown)


src/bandersnatch/tests/plugins/test_blocklist_name.py
```diff
@@ -121,10 +121,41 @@ def test__filter__name_only(self) -> None:
         )
 
         mirror = Mirror(Path("."), Master(url="https://foo.bar.com"))
-        mirror.packages_to_sync = {"foo": "4.5.6"}
+        mirror.packages_to_sync = {"foo": "", "foo==1.2.3": ""}
```
cooperlees, line 124: As disucssed, can we please just remove this confusing test cases since xmlrpc can't return something like this.

PR #678, opened 2020-08-30
Simplify logging around filters

dralley:


src/bandersnatch/mirror.py
```diff
@@ -78,6 +78,13 @@ async def synchronize(
                 for name in specific_packages
             }
 
+        if not self.filters.filter_metadata_plugins:
+            logger.info("No metadata filters are enabled. Skipping metadata filtering")
+        if not self.filters.filter_release_plugins:
+            logger.info("No release filters are enabled. Skipping release filtering")
+        if not self.filters.filter_release_file_plugins:
+            logger.info("No release file filters are enabled. Skipping release file filtering")
+
```
dralley, line 87: There should be only two semantic differences:

1. The origin of the message that is logged will be different, it'll come from mirror instead of package.  I don't see that as making a difference.
2.  Before, these 3 were sharing the same global variable, so I think once one message was logged, the others would *not* be logged, even though the messages are different and mean different things.  That feels erroneous to me

dralley:
Looks like PyPI is having issues again : /

src/bandersnatch/mirror.py
```diff
@@ -42,9 +42,7 @@ class Mirror:
     now = None
 
     def __init__(
-        self,
-        master: Master,
-        workers: int = 3,
+        self, master: Master, workers: int = 3,
```
cooperlees, line 45: You have to remove the trailing ',' with latest black if you want this.

src/bandersnatch/mirror.py
cooperlees: Agree - As long as we're not not spamming the message for each package and we only say it once.

src/bandersnatch/mirror.py
```diff
@@ -297,6 +303,9 @@ async def process_package(self, package: Package) -> None:
             return None
 
         # save the metadata before filtering releases
+        # (dalley): why? the original author does not remember, and it doesn't seem
+        # to make a lot of sense.
+        # https://github.com/pypa/bandersnatch/commit/2a8cf8441b97f28eb817042a65a042d680fa527e#r39676370
         if self.json_save:
```
cooperlees, line 309: I agree we should only save JSON for mirror'ed packages based on filters. Happy for that PR. Would prefer it a dedicated PR tho.

src/bandersnatch/mirror.py
dralley: Yeah this is just an artifact from running black locally, I don't have the latest version apparently.

src/bandersnatch/mirror.py
dralley: Yup, that would be no problem.  I just wanted to document first in case we ever ran across a reason to do so.

dralley:
@cooperlees It's actually not logging per-package, it's just logging once at the beginning of the sync process.

Like I mentioned earlier the messages are similar, so I can see how at a quick glance they could be mistaken for the same message being printed multiple times, but it's not the case :)

```
 2020-08-30 20:41:18,046 INFO: 4 packages to sync.
INFO:bandersnatch.mirror:4 packages to sync.
2020-08-30 20:41:18,054 INFO: No metadata filters are enabled. Skipping metadata filtering
INFO:bandersnatch.mirror:No metadata filters are enabled. Skipping metadata filtering
2020-08-30 20:41:18,054 INFO: No release filters are enabled. Skipping release filtering
INFO:bandersnatch.mirror:No release filters are enabled. Skipping release filtering
2020-08-30 20:41:18,054 INFO: No release file filters are enabled. Skipping release file filtering
INFO:bandersnatch.mirror:No release file filters are enabled. Skipping release file filtering
```

Before, it would have just printed 

```
2020-08-30 20:41:18,054 INFO: No metadata filters are enabled. Skipping metadata filtering
INFO:bandersnatch.mirror:No metadata filters are enabled. Skipping metadata filtering
```

because of the bool sharing

cooperlees:
Wonder why we get the two lines there. I won't block on that, but would love to root cause and squash that bug. Rebasing.

PR #699, opened 2020-09-26
swift.py: use BaseFileLock's lock_file property

hauntsaninja:
filelock.BaseFileLock has had this property for forever: https://github.com/benediktschmitt/py-filelock/blob/b30bdc4fb998f5f4350257235eb51147f9e81862/filelock.py#L159. Using it seems better than using the private attribute.

https://github.com/python/typeshed/pull/4424 added stubs for filelock to typeshed, which don't have the private attribute. This means an upcoming release of mypy would complain about these lines.


cooperlees:
I'm sorry here - But I have very little Swift knowledge + @techalchemy has not been around of late.

If you rebase I'll merge this as it makes sense to me!

cooperlees:
I've tried a branch update here, if CI passes I'll merge, but we don't have good swift testing.

cooperlees:
Merged. Thanks! If you could test in your prod use and if any bugs send em up that would be appreciated.

PR #726, opened 2020-10-29
Fixes PYPI_SERIAL_HEADER

liyingjun:
PYPI_SERIAL_HEADER should be 'X-PyPI-Last-Serial' instead of
'X-PYPI-LAST-SERIAL'. Example below:

curl -IL https://pypi.org/pypi/pip/json
...
X-PyPI-Last-Serial: 8436753
...


cooperlees:
HTTP headers (according to RFC2616) are case insensitive. What is this actually breaking for you here?

cooperlees:
Please re-open if you plan to explain what this breaks + rebase etc.

Thanks for offering a fix tho!

PR #738, opened 2020-11-19
now slimmer docker image

nlaurance-pyie:
By building the dependencies beforehand and reinstall on a slim version

weighting/bandersnatch   builder             503cec72f6a7        12 seconds ago      279MB
weighting/bandersnatch   base                0c7eb5b38afb        56 minutes ago      1.05GB

resulting image is +/- 4x smaller

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3 as base
+FROM python:3.9 as base
 
 FROM base as builder
 RUN mkdir /install
@@ -11,7 +11,7 @@ RUN pip install --target="/install" \
     -r requirements_swift.txt
 
 
-FROM python:3-slim
+FROM python:3.9-slim
 
 COPY --from=builder /install /usr/local/lib/python3.9/site-packages
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,22 +1,28 @@
-FROM python:3
+FROM python:3 as base
 
-RUN mkdir /bandersnatch
-RUN mkdir /conf && chmod 777 /conf
-ADD setup.cfg /bandersnatch
-ADD setup.py /bandersnatch
-ADD requirements.txt /bandersnatch
-ADD requirements_swift.txt /bandersnatch
-ADD README.md /bandersnatch
-ADD CHANGES.md /bandersnatch
+FROM base as builder
+RUN mkdir /install
+WORKDIR /install
+ADD requirements_swift.txt /install
+ADD requirements.txt /install
+RUN pip install --target="/install" --upgrade pip setuptools wheel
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,22 +1,28 @@
-FROM python:3
+FROM python:3.9 as base
```
cooperlees, line 1:
```suggestion
FROM python:3 as base
```

Does this work so we move with python versions?

Dockerfile
```diff
@@ -1,22 +1,28 @@
...
+WORKDIR /install
+ADD requirements_swift.txt /install
+ADD requirements.txt /install
+RUN pip install --target="/install" --upgrade pip setuptools wheel
+RUN pip install --target="/install" \
+    -r requirements.txt \
+    -r requirements_swift.txt
+
+
+FROM python:3.9-slim
```
cooperlees, line 14: Again, I'd rather not need to update there ...

Dockerfile
```diff
@@ -1,22 +1,28 @@
...
 
-# OPTIONAL: Include a config file
-# Remember to bind mount the "directory" in bandersnatch.conf
-# Reccomended to bind mount /conf - `runner.py` defaults to look for /conf/bandersnatch.conf
-# ADD bandersnatch.conf /etc
+COPY --from=builder /install /usr/local/lib/python3.9/site-packages
```
cooperlees, line 16: Is there some dynamic way so we can no hard code to 3.9?

cooperlees (changes requested):
Many thanks for this. I need to learn about this FROM etc. so I can make many of my docker images smaller.

Is there anyways we can not hard code to 3.9? I like to be lazy and only do work when CI fails.

Dockerfile
nlaurance-pyie: I'd rather be explicit about the version used

Dockerfile
nlaurance-pyie: don't think so, main reason version is pinned up there
I'm happy to be explicit, not too much work to tune to 3.8 or 3.10  :)

cooperlees: Yeah, it will just lag, we'll forget / miss it when next version comes out etc. I try to avoid that.

If we can't I'll open an issue to look into it and merge for now as the size wins are worth the work.

Dockerfile
cooperlees: Well, we build every commit so my theory is that people can choose the docker container built on the python version they want if it matter.

My other theory is to just keep bandersnatch working on $latest python3 tag.

cooperlees (approved):
Will accept and see if we can move back to python3 one day.

PR #746, opened 2020-11-24
Setup dependabot so Actions are automatically updated

ichard26:
One thing less to remember to update :D

Implemented according to https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/keeping-your-actions-up-to-date-with-dependabot

FYI a `dependencies` label [will be created automatically for use by Dependabot](https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/configuration-options-for-dependency-updates#labels). Unfortunately it appears there is no way to disable this behaviour. I guess configure PyUP to add that label too now?

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,9 @@
+version: 2
+updates:
+  # Enable version updates for Actions used in GitHub Actions workflows
+  - package-ecosystem: "github-actions"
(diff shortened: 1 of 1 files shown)


cooperlees (approved):
O hell yes. I've been waiting for this - Will have to add it to all my projects! Thanks.

Since this can do python + github actions I might retire pyup.io.

PR #761, opened 2020-12-10
Fixed #760. Logging double output.

### LINKED ISSUE #760: Logging broken; produces double output
Logging is currently broken:
```
2020-12-10 18:06:46,738 INFO: Syncing with https://pypi.org.
INFO:bandersnatch.mirror:Syncing with https://pypi.org.
2020-12-10 18:06:46,738 INFO: Current mirror serial: 8869123
INFO:bandersnatch.mirror:Current mirror serial: 8869123
```
Issue is due to bad call to logging module. Will supply a PR.

electricworry:


cooperlees (approved):
This causes every logging message to double write across the whole program? Weird.

electricworry:
> This causes every logging message to double write across the whole program? Weird.

Indeed. I don't fully understand it myself and took me a while to track down. But switching that one log output from logging to the logger object fixed it for me.

electricworry:
I think I found the [explanation](https://docs.python.org/2/library/logging.html#logging.log):

> The above module-level convenience functions, which delegate to the root logger, call basicConfig() to ensure that at least one handler is available.

This will result in another handler in the logger chain, I think.

cooperlees:
Thanks - I don't get the coverage bot. Ignoring that.

PR #763, opened 2020-12-11
Handle 404 status for json verify

electricworry:
Gracefully handle 404 response for verify --json-update. If --delete also specified, old json deleted, which leads to its disowned packages being deleted.

### DIFF
--- a/src/bandersnatch/verify.py
+++ b/src/bandersnatch/verify.py
@@ -12,6 +12,7 @@ from pathlib import Path
 from sys import stderr
 from typing import List, Optional, Set
 from urllib.parse import urlparse
+import aiohttp
 
 from .filter import LoadedFilters
 from .master import Master
--- a/src/bandersnatch/verify.py
+++ b/src/bandersnatch/verify.py
@@ -12,6 +12,7 @@ from pathlib import Path
 from sys import stderr
 from typing import List, Optional, Set
 from urllib.parse import urlparse
+
 import aiohttp
 
 from .filter import LoadedFilters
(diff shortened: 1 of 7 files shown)


cooperlees (approved):
Makes sense, if you want bonus point could mock / monkey patch url_fetch to raise a ClientResponseError exception to test the code. All good if you don't have time.

electricworry:
I'll take a look at that. Do I just update this branch and it will automatically get added to the PR?

cooperlees:
> I'll take a look at that. Do I just update this branch and it will automatically get added to the PR?

Yup, sure will. Just `git push origin verify-404`

cooperlees:
@electricworry - isort fails in link. I haven't debugged yet but here is the error:
- https://pastebin.com/Q0twYNg5

ichard26:
The `W0503: Deprecated` message is dealt with in PR #767 :D

src/bandersnatch/verify.py
```diff
@@ -12,6 +12,7 @@
 from sys import stderr
 from typing import List, Optional, Set
 from urllib.parse import urlparse
+import aiohttp
```
cooperlees, line 15: Pretty sure this just needs to move to make isort happy now.

cooperlees (changes requested):
Please just run tox lint which runs pre-commit and it should make isort happy.

src/bandersnatch/verify.py
electricworry: Done with "tox -e lint". Got there in the end. Been a busy few days.

cooperlees:
No worries. Thanks for the fix and learning our testing!

Merging as we know the coverage is missed.

PR #776, opened 2020-12-20
Fix links, improve docs CI, and improve external object linking

ichard26:
* Fix broken or redirecting links

* More intersphinx linking
    
    Now references to packaging and filelock objects will be linked to their
    external documentation.
    
    Also removed urllib3 because no where does our code or docs reference it.

* Add linkcheck to tox config

    This commit exists because I didn't want to have to remember to run
    linkcheck routinely. Call me lazy if you will :P ... Well actually,
    running linkcheck on CI via tox is a good way to make sure links
    that go stale are noticed and fixed quickly.

    I added --keep-going because it's better to get all of the warnings
    in a single run then error out the build *than* only knowing that
    there's at least one warning somewhere. Very useful if you are
    looking at the CI logs as you can't just ask logs to tell what all the
    warnings were about ;)

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -14,7 +14,8 @@ extras = swift
 [testenv:doc_build]
 basepython=python3
 commands =
-    {envpython} {envbindir}/sphinx-build -a -W -b html docs docs/html
+    {envpython} {envbindir}/sphinx-build -a -W --keep-going -b html docs docs/html
+    {envpython} {envbindir}/sphinx-build -a -W --keep-going -b linkcheck docs docs/html
 changedir = {toxinidir}
 deps =
     -r requirements_docs.txt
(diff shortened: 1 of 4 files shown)


ichard26:
Hm, I should probably add a step to CI that runs `linkcheck` so I don't have to remember to run this.

tox.ini
```diff
@@ -14,7 +14,8 @@ extras = swift
 [testenv:doc_build]
 basepython=python3
 commands =
-    {envpython} {envbindir}/sphinx-build -a -W -b html docs docs/html
+    {envpython} {envbindir}/sphinx-build -a -W --keep-going -b html docs docs/html
+    {envpython} {envbindir}/sphinx-build -a -W --keep-going -b linkcheck docs docs/html
```
cooperlees, line 18: This is not lazy this is smart.

cooperlees (approved):
Awesome. Thanks @ichard26 !

tox.ini
ichard26: I thought it was clear enough I was joking :sweat_smile:

PR #785, opened 2020-12-30
Handling of timeouts that can occur in verify.

electricworry:
Also added tests for comment.

### DIFF
--- a/src/bandersnatch/tests/test_mirror.py
+++ b/src/bandersnatch/tests/test_mirror.py
@@ -1,4 +1,5 @@
 import os.path
+import sys
 import unittest.mock as mock
 from os import sep
 from pathlib import Path
@@ -1063,3 +1064,7 @@ def test_determine_packages_to_sync(mirror: BandersnatchMirror) -> None:
     mirror.packages_to_sync = {"black": 69, "foobar": 47, "barfoo": 68}
     target_serial = mirror.find_target_serial()
     assert target_serial == 69
+
+
+if __name__ == "__main__":
+    pytest.main(sys.argv)
--- a/src/bandersnatch/tests/test_verify.py
+++ b/src/bandersnatch/tests/test_verify.py
@@ -33,13 +33,6 @@ def some_dirs(*args: Any, **kwargs: Any) -> List[str]:
     return ["/data/pypi/web/json/bandersnatch", "/data/pypi/web/json/black"]
 
 
-class FakeArgs:
-    delete = True
-    dry_run = False
-    json_update = True
-    workers = 2
-
-
 class FakeConfig:
     def get(self, section: str, item: str) -> str:
         if section == "mirror":
@@ -206,6 +199,11 @@ async def test_get_latest_json(monkeypatch: MonkeyPatch) -> None:
 
 @pytest.mark.asyncio
 async def test_metadata_verify(monkeypatch: MonkeyPatch) -> None:
+    class FakeArgs:
+        delete = True
+        dry_run = True
+        workers = 2
+
     fa = FakeArgs()
     fc = FakeConfig()
     monkeypatch.setattr(bandersnatch.verify, "verify_producer", do_nothing)
@@ -215,11 +213,15 @@ async def test_metadata_verify(monkeypatch: MonkeyPatch) -> None:
 
 
 @pytest.mark.asyncio
-async def test_get_latest_json_timeout(
-    monkeypatch: MonkeyPatch, tmp_path: Path
-) -> None:
-    fc = FakeConfig()
+async def test_get_latest_json_timeout(tmp_path: Path) -> None:
+    class FakeArgs:
+        delete = True
+        dry_run = False
+        json_update = True
+        workers = 2
+
     fa = FakeArgs()
+    fc = FakeConfig()
 
     master = Master(fc.get("mirror", "master"))
     url_fetch_timeout = mock.AsyncMock(side_effect=ServerTimeoutError)  # type: ignore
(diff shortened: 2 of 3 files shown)


src/bandersnatch/tests/test_mirror.py
```diff
@@ -1063,3 +1064,7 @@ def test_determine_packages_to_sync(mirror: BandersnatchMirror) -> None:
     mirror.packages_to_sync = {"black": 69, "foobar": 47, "barfoo": 68}
     target_serial = mirror.find_target_serial()
     assert target_serial == 69
+
+
+if __name__ == "__main__":
+    pytest.main(sys.argv)
```
cooperlees, line 1070: Was this just so you could run it directly to test? Can also do that with `pytest` and some args I always forget.

src/bandersnatch/tests/test_verify.py
```diff
@@ -31,24 +33,21 @@ def some_dirs(*args: Any, **kwargs: Any) -> List[str]:
...
-
-
 class FakeConfig:
     def get(self, section: str, item: str) -> str:
         if section == "mirror":
             if item == "directory":
                 return "/data/pypi"
             if item == "master":
-                return "https://pypi.org/simple/"
+                return "https://unittest.org"
```
cooperlees, line 42: Agree - def makes errors clearer. Thanks.

src/bandersnatch/tests/test_verify.py
```diff
@@ -31,24 +33,21 @@ def some_dirs(*args: Any, **kwargs: Any) -> List[str]:
...
         return ""
 
     def getfloat(self, section: str, item: str, fallback: float = 0.5) -> float:
         return 0.5
 
+    def getint(self, section: str, item: str, fallback: int = 5) -> float:
+        return 5
```
cooperlees, line 49: Maybe I / someone also missed we should use fallback ?
```suggestion
        return fallback

    def getint(self, section: str, item: str, fallback: int = 5) -> float:
        return fallback
```

src/bandersnatch/verify.py
```diff
@@ -109,7 +109,15 @@ async def verify(
 
     if args.json_update:
         if not args.dry_run:
-            await get_latest_json(master, json_full_path, config, executor, args.delete)
+            try:
+                await get_latest_json(
+                    master, json_full_path, config, executor, args.delete
+                )
+            except Exception:
```
cooperlees, line 116: I always recommend (and it's a good practice) catching only the exceptions you expect explicitly. Here I feel it's better to error then keep on going if someone wants new JSON to be downloaded before trying to verify a package.

Do we actually want to continue here? If so, I feel we should respect the config `stop-on-error`

Maybe This?
```suggestion
            except aiohttp.ClientError:
```

src/bandersnatch/verify.py
```diff
@@ -137,7 +145,14 @@ async def verify(
                     all_package_files.append(pkg_file)
                     continue
                 else:
-                    await master.url_fetch(jpkg["url"], pkg_file, executor)
+                    try:
+                        await master.url_fetch(jpkg["url"], pkg_file, executor)
+                    except Exception:
```
cooperlees, line 150: Again only really expecting http timeouts (sure we could get OS Errors) ...
```suggestion
                    except aiohttp.ClientError:
```

src/bandersnatch/tests/test_verify.py
```diff
@@ -208,5 +212,90 @@ async def test_metadata_verify(monkeypatch: MonkeyPatch) -> None:
...
+        delete = True
+        dry_run = False
+        json_update = True
+        workers = 2
+
+    fa = FakeArgs()
+    fc = FakeConfig()
+
+    master = Master(fc.get("mirror", "master"))
+    url_fetch_404 = mock.AsyncMock(  # type: ignore
```
cooperlees, line 253: I think typing actually helped here - AsyncMock is not available to 3.8 and we still support 3.6.1 ... Maybe now could be the time to go to >= 3.8 ...
```suggestion
    url_fetch_404 = mock.AsyncMock(  # type: ignore
```

cooperlees (changes requested):
Thanks! This is awesome and great testing.

AsyncMock is >= 3.8 tho. You'll see I am using asynctest in other tests due to this. Can you use this and we'll do another PR after my next release moving us to >= 3.8. I think it's time.

src/bandersnatch/tests/test_mirror.py
electricworry: Yes, and just making it consistent with test_verify.py. I suppose all test files should have it present or not consistently.

src/bandersnatch/verify.py
electricworry: I generally agree. But I was basing this on the general behaviour of the mirror mode which catches all Exceptions to allow the command to continue. Also, it's not just client error exceptions but also timeouts that I was experiencing, and these would cause the process to stop completely, so just took the approach of log and continue.

src/bandersnatch/tests/test_verify.py
electricworry: Sorry. Happy to change if not ready for >= 3.8. Will await decision.

src/bandersnatch/verify.py
electricworry: Same as above.

src/bandersnatch/tests/test_verify.py
electricworry: This change would affect, for example LTS users of Ubutnu 18.04 and 16.04 wouldn't it?

ichard26: Yeah, Ubuntu 18.04 LTS only has CPython 3.6.something pre-installed and Ubuntu 16.04 LTS only has CPython 3.5.something pre-installed. (I know CPython 2.7.something is also included on both but let's just ignore them)

cooperlees: bandersnatch has ignored python2 since I took over :D My first contributions were moving from Python2 only to Python3 :)

cooperlees (approved):
LGTM - Thanks!

In future if you can remember the CHANGELOG that saves me time. But I forgot to ask so I'll add once I release.

electricworry:
Sorry, it's an obvious thing. I should have noticed the changelog and prepended. I just took a look at CONTRIBUTING.md and I don't think there's an explicit mention of it. Think I might add mention to it for idiots like me. :)

Thanks for accepting this. I've got some improvements and fixes I want to promote and I'd like to expand on the documentation. I've got a bit of time now.

PR #795, opened 2021-01-08
Make RTD aware of our love (need) for Python 3.8

ichard26:
Lemme test this on my bandersnatch ReadTheDocs setup first. 

Also ReadTheDocs doesn't support 3.9 :(

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -16,7 +16,7 @@ formats:
   - epub
 
 python:
-  version: 3.7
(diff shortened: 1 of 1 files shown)


cooperlees (approved):
3.9 will come. I will see if we can remove this version IF readthedocs goes to latest version by default. I'd prefer this rather than us hard setting unless we have to.

3.9 Issue: https://github.com/readthedocs/readthedocs.org/issues/7554

PR #798, opened 2021-01-10
Update / cleanup stuff relating to Python 3.7 and below

ichard26:
A PR fully created with the GitHub Web UI, aww yea!

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -3,17 +3,15 @@
 The following instructions will place the bandersnatch executable in a
 virtualenv under `bandersnatch/bin/bandersnatch`.
 
-- bandersnatch **requires** `>= Python 3.6`
+- bandersnatch **requires** `>= Python 3.8.0`
 
 
 ### pip
 
 This installs the latest stable, released version.
 
-*(>= 3.6.1 required)*
-
 ```
-  $ python3.6 -m venv bandersnatch
+  $ python3.8 -m venv bandersnatch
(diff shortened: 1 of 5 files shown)


ichard26:
Dammit using the web UI was a bad idea. On the upside we get to say goodbye to a few type ignore comments :)

ichard26:
Actually I found even more stuff related to Python 3.7 and below .... ahhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhh. Give me a moment!

ichard26:
Cool, that should be all of them, except for one of them since changing it probably affects how bandersnatch can be called:

https://github.com/pypa/bandersnatch/blob/c3673b9024bec0b201788ec9cfa28af81b840017/src/bandersnatch/main.py#L203-L209

`asyncio.run` doesn't support using an already running event loop if I can read docs properly.

cooperlees:
> Cool, that should be all of them, except for one of them since changing it probably affects how bandersnatch can be called:
> 
> https://github.com/pypa/bandersnatch/blob/c3673b9024bec0b201788ec9cfa28af81b840017/src/bandersnatch/main.py#L203-L209
> 
> `asyncio.run` doesn't support using an already running event loop if I can read docs properly.

If people are calling bandersnatch I doubt they are using main.py. If they are we have some refactoring to do.

Thanks @ichard26! Sometimes I question keeping the coverage bot ... all open to suggestions there.

cooperlees (approved):

PR #799, opened 2021-01-10
Check if we need to actually attempt to load plugins

### LINKED ISSUE #649: Remove `keystoneauth1` error when using filesystem storage
```
2020-08-10 00:12:50,875 INFO: Selected storage backend: filesystem
ERROR:root:Unable to load entry point swift_plugin = bandersnatch_storage_plugins.swift:SwiftStorage: No module named 'keystoneauth1'
```
- Example CI Job I saw this in: https://github.com/pypa/bandersnatch/pull/648/checks?check_run_id=964744678

I think this is due to not installing the deps, but why do we need to load this if we're not using it? Is this a limitation of the way plugins are? How could we refactor if so?

electricworry:
Fixes #649

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -12,4 +12,6 @@ exclude =
     __pycache__,
     *.pyc,
     .git,
-    .tox
+    .tox,
+    venv,
+    ENV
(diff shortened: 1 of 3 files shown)


cooperlees (approved):
Thanks. LGTM. Just made the PR title + summary a little more complete

Issue #802, opened 2021-01-14
GitHub Actions Docker Upload is flaky


cooperlees:
Seems connection timeouts with pip from the few I've looked at. Would love some help if anyone has time to workout why the uploads are failing. Seems building the swift image in particular.

Example Fail: https://github.com/pypa/bandersnatch/runs/1697632568?check_suite_focus=true

ichard26:
This is because of Pip's new dependency resolver that is a lot stricter when resolving dependencies. Unfortunately being strict means reading a bunch of metadata which is expensive since whole files have to be downloaded. Looks like we have a complex dependency tree situation on our hands...

PR #803, opened 2021-01-15
Ban idna 3.x series due to compatibility issues

ichard26:
Fixes https://github.com/pypa/bandersnatch/issues/802

idna 3 and above causes some extreme backtracking because
requests doesn't support the 3.x series. Requests is
heavily used by dependencies needed for bandersnatch
swift support so when requests must downgrade to a
version that vendors idna, dependency tree turns into a
variant that pip's new dependency resolver doesn't like
resolving. Although not installing the swift dependencies
or using the unpinned dependencies info in setup.cfg
works properly.

This also fixes the long pip install process which caused certain GHA
jobs to timeout.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,16 @@ async-timeout==3.0.1
 attrs==20.3.0
 chardet==3.0.4  # pyup: ignore
 filelock==3.0.12
-idna==3.1
+idna==2.10  # pyup: ignore
+            # idna 3 and above causes some extreme backtracking because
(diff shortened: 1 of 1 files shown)


cooperlees:
Many thanks for digging this out @ichard26. Just out of curiosity, what else depends on idna other than requests? Anything?

cooperlees (approved):

ichard26:
>  Just out of curiosity, what else depends on idna other than requests?

`yarl` which is needed by `aiohttp`.

PR #806, opened 2021-01-18
Implement a new filter that denied/allows projects based on total filesize.

electricworry:
WIP. Functionally works, but need to document, add tests, and make some improvements. Also fixed some issues spotted on the way.

### DIFF
--- a/src/bandersnatch/mirror.py
+++ b/src/bandersnatch/mirror.py
@@ -211,7 +211,6 @@ class BandersnatchMirror(Mirror):
         self.homedir = self.storage_backend.PATH_BACKEND(str(homedir))
         self.lockfile_path = self.homedir / ".lock"
         self.master = master
-        self.filters = LoadedFilters(load_all=True)
 
         # Stop soon after meeting an error. Continue without updating the
         # mirror's serial if false.
--- a/src/bandersnatch/tests/plugins/test_metadata_plugins.py
+++ b/src/bandersnatch/tests/plugins/test_metadata_plugins.py
@@ -1,14 +1,16 @@
-from unittest import TestCase
-from tempfile import TemporaryDirectory
 import os
-import bandersnatch.filter
+from pathlib import Path
+from tempfile import TemporaryDirectory
+from typing import cast
+from unittest import TestCase
 
 from mock_config import mock_config
-from pathlib import Path
 
-from bandersnatch.mirror import BandersnatchMirror
+import bandersnatch.filter
 from bandersnatch.master import Master
+from bandersnatch.mirror import BandersnatchMirror
 from bandersnatch.package import Package
+from bandersnatch_filter_plugins.metadata_filter import SizeProjectMetadataFilter
 
 
 class TestSizeProjectMetadataFilter(TestCase):
@@ -43,7 +45,9 @@ max_package_size = 1G
         names = [plugin.name for plugin in plugins]
         self.assertListEqual(names, ["size_project_metadata"])
         self.assertEqual(len(plugins), 1)
-        self.assertTrue(plugins[0].initialized)
+        self.assertIsInstance(plugins[0], SizeProjectMetadataFilter)
+        plugin: SizeProjectMetadataFilter = cast(SizeProjectMetadataFilter, plugins[0])
+        self.assertTrue(plugin.initialized)
 
     def test__filter__size__only(self) -> None:
         mock_config(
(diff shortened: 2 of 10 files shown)


cooperlees (commented):
Good to see someone else makes consistent spelling mistakes in code like me (or was it me?)

Looks good, but I am sure you plan it, but please:
- add a quick test over the filter() function at minimum

For some bonus points:
- Please add CHANGES entry
- Please add config + explanation of what it does to documentation

electricworry:
100%. Documentation incoming in just a moment, tests a little later.
I also want to implement human readable size values, e.g. "1G" instead of "1000000000". Do you know of a good quality package that does this conversion already? (from human to int.)

cooperlees:
> Do you know of a good quality package that does this conversion already? (from human to int.)
Maybe https://pypi.org/project/humanfriendly/

src/bandersnatch/tests/plugins/test_metadata_plugins.py
```diff
@@ -0,0 +1,130 @@
...
+    def setUp(self) -> None:
+        self.cwd = os.getcwd()
+        self.tempdir = TemporaryDirectory()
+        os.chdir(self.tempdir.name)
+
+    def tearDown(self) -> None:
+        if self.tempdir:
+            assert self.cwd
+            os.chdir(self.cwd)
+            self.tempdir.cleanup()
```
cooperlees, line 30: Do your tests actually write out any files? I never understood if these are actually needed. I'm guessing you just copy pasted from other test, but was just wondering. I guess it will write out some base mirror directory structure.

src/bandersnatch_filter_plugins/allowlist_name.py
```diff
@@ -64,7 +64,7 @@ def filter(self, metadata: Dict) -> bool:
 
     def check_match(self, **kwargs: Any) -> bool:
         """
-        Check if the package name matches against a project that is blocklisted
```
cooperlees, line 67: Nice catch.

src/bandersnatch_filter_plugins/metadata_filter.py
```diff
@@ -176,14 +179,83 @@ def filter(self, metadata: Dict) -> bool:
...
+                )
+                return
+            if self.max_package_size > 0:
+                if not self.allowlist_package_names:
+                    self.allowlist_package_names = (
+                        self._determine_unfiltered_package_names()
+                    )
+
+                logger.info(
+                    f"Initialized metadata plugin {self.name} to block projects >{self.max_package_size}b"  # noqa: E501
```
cooperlees, line 223: I guess the E501 is to keep your local editor happy?

src/bandersnatch/mirror.py
```diff
@@ -211,7 +211,6 @@ def __init__(
         self.homedir = self.storage_backend.PATH_BACKEND(str(homedir))
         self.lockfile_path = self.homedir / ".lock"
         self.master = master
-        self.filters = LoadedFilters(load_all=True)
```
cooperlees, line 214: This seems a big change. Why? Can you explain how the functionality has been retained from removing this please or how it's was not needed.

cooperlees (changes requested):
Looks great. Just one question about the change to mirror.py - Just want an explanation to understand and make sure it's not breaking anyting.
- CI does agree the `allowlist_project` plugin is working tho as it uses it ...

src/bandersnatch/tests/plugins/test_metadata_plugins.py
electricworry: it's true. I'm guilty of copy/pasting. I just presumed that it was a required practice I should follow. :)

src/bandersnatch/mirror.py
electricworry: I should probably not have done this as part of an enhancement, but it's a superflous line. It's set in the superclass's constructor.

src/bandersnatch_filter_plugins/metadata_filter.py
electricworry: Just a line too long. I'll split it if desired, but I thought it might get a bit confusing with the ternary if-else following.

src/bandersnatch/tests/plugins/test_metadata_plugins.py
ichard26: And [to some](url), not best practice-y enough since `tearDown()` won't be called if `setUp()` fails.

cooperlees: pytest is lots of voodoo. Maybe we can make an issue to test / clean these up.

cooperlees (approved):
All makes sense to me now. Thanks!

Issue #821, opened 2021-02-04
Enhancement: check file size and modification time(upload time) in download_file


jiegec:
Relevant code: https://github.com/pypa/bandersnatch/blob/cc50bfecfd8b8912c35227aa8124c385ca1239ab/src/bandersnatch/mirror.py#L771-L786

It reads the local file and calculates sha256sum. This might lead to huge pressure to disk. An enhancement would be:

1. check file size and modification time(upload time) by default
2. check sha256sum if user explicitly requests in config

cooperlees:
Thanks for the idea and the PR. As stated on the PR, I want to keep the checksumming as the default and add this as a configuration option. If you implement that I'll merge away!

cooperlees:
This is merged ready for 5.0. Thanks @jiegec !

PR #822, opened 2021-02-04
Skip downloading based on file size and upload time instead of sha256sum

jiegec:
See issue #821. This pr implements:

1. check file size and modification time(upload time)

It currently does not allow user to force sha256sum check. And this pr is tested only on local filesystems, but not OpenStack.

### DIFF
--- a/src/bandersnatch/mirror.py
+++ b/src/bandersnatch/mirror.py
@@ -198,6 +198,7 @@ class BandersnatchMirror(Mirror):
         *,
         cleanup: bool = False,
         release_files_save: bool = True,
+        compare_method: Optional[str] = None,
     ) -> None:
         super().__init__(master=master, workers=workers)
         self.cleanup = cleanup
@@ -230,6 +231,7 @@ class BandersnatchMirror(Mirror):
         self.diff_full_path = diff_full_path
         self.keep_index_versions = keep_index_versions
         self.digest_name = digest_name if digest_name else "sha256"
+        self.compare_method = compare_method if compare_method else "hash"
         self.workers = workers
         self.diff_file_list = diff_file_list or []
         if self.workers > 10:
@@ -640,7 +642,8 @@ class BandersnatchMirror(Mirror):
         for release_file in package.release_files:
             try:
                 downloaded_file = await self.download_file(
-                    release_file["url"], release_file["digests"]["sha256"]
+                    release_file["url"], release_file["digests"]["sha256"],
+                    release_file["size"]
                 )
                 if downloaded_file:
                     downloaded_files.add(str(downloaded_file.relative_to(self.homedir)))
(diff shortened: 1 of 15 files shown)


src/bandersnatch/mirror.py
```diff
@@ -640,7 +641,10 @@ async def sync_release_files(self, package: Package) -> None:
         for release_file in package.release_files:
             try:
                 downloaded_file = await self.download_file(
-                    release_file["url"], release_file["digests"]["sha256"]
+                    release_file["url"],
+                    release_file["size"],
+                    isoparse(release_file["upload_time_iso_8601"]),
```
cooperlees, line 646: I'm a big fan of keeping dependencies down. I feel native datetime can parse this from strftime - Can we do that to drop the dependency please?

src/bandersnatch_storage_plugins/filesystem.py
```diff
@@ -262,3 +263,22 @@ def get_hash(self, path: PATH_TYPES, function: str = "sha256") -> str:
...
+        return datetime.datetime.fromtimestamp(
+            path.stat().st_mtime, datetime.timezone.utc
+        )
+
+    def set_upload_time(self, path: PATH_TYPES, time: datetime.datetime) -> None:
+        """Set the upload time of a given **path**"""
+        if not isinstance(path, pathlib.Path):
+            path = pathlib.Path(path)
+        ts = time.timestamp()
+        os.utime(path, (ts, ts))
```
cooperlees, line 284: Could the file ever not exist? Should we check for that or do we handle it catching the exceptions in the callers?

cooperlees (changes requested):
Thanks for this. "Checksumming" is vastly superior to just checking the size + date, but I do agree the combination of size + time is enough entropy. I feel I can only accept this if you make it config driven, with the check summing remaining the default.

I also REALLY appreciate the testing + adding in of swift calls!

Can you please:
- Add a config option to enable / disable file size + mtime checking
  - Maybe `size_date_validate` or something better
- Document it in /docs/configuration.md please
- Add a CHANGES.md entry for this please

What was the speedup from doing this? Do you have any numbers on IO savings or CPU / Real time saved?

jiegec:
Working on other projects recently.. Will come back to finish this pr later.

src/bandersnatch/mirror.py
jiegec: This is done in https://github.com/pypa/bandersnatch/pull/822/commits/880650fd9f156ebe83a8694031e92da86410b2a6#diff-8b844ef0ea8cbe96fbc0bd8fdfebc3b9e245509c5909654721be845486d0f82fR647

src/bandersnatch_storage_plugins/filesystem.py
happyaron: At the moment, the only caller of these functions is `download_file()` in _mirror.py_, the `get_*()` ones are only called within an `if path.exist()` block, and `set_upload_time()`'s usage is extended to a situation right after a download is successful. If the file is gone afterwards then it signals something wrong is happening.

cooperlees (approved):
Thanks! This is great and the main parts are all tested! Nice addition if people want to have this less IO intensive stat checking.

Welcome to the Bandersnatch contributors family! This will go out to PyPI when I get 5.0 sorted.

PR #853, opened 2021-03-14
Add helpful echo to direct people to add changes 

cooperlees:
- When news / changelog CI fails have a blatant echo to say "Add the line" to try help people not get confused

cc: @ichard26

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -17,4 +17,5 @@ jobs:
       if: contains(github.event.pull_request.labels.*.name, 'skip news') != true
       run: |
         grep -P "PR #${{ github.event.pull_request.number }}[^0-9]" CHANGES.md || \
-        echo "Please add 'PR #${{ github.event.pull_request.number }}' change line to CHANGES.md"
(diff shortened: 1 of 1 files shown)


.github/workflows/changelog.yml
```diff
@@ -16,4 +16,5 @@ jobs:
     - name: Grep CHANGES.md for PR number
       if: contains(github.event.pull_request.labels.*.name, 'skip news') != true
       run: |
-        grep -P "PR #${{ github.event.pull_request.number }}[^0-9]" CHANGES.md
+        grep -P "PR #${{ github.event.pull_request.number }}[^0-9]" CHANGES.md || \
+        echo "Please add 'PR #${{ github.event.pull_request.number }}' change line to CHANGES.md"
```
cooperlees, line 20: We might need a `; exit 1` after the echo so the job fails

ichard26: For future reference, a `; exit 1` was necessary to fail the job after the echo.

ichard26 (approved):

Issue #787, opened 2020-12-31
Replace asynctest (Requires >=3.8)


cooperlees:
Lets see if we can drop asynctest as it has outdated asyncio code etc. and seems unmaintained.

It's using old deprecated code:
```
/home/runner/work/bandersnatch/bandersnatch/.tox/py3/lib/python3.9/site-packages/asynctest/mock.py:434: DeprecationWarning: "@coroutine" decorator is deprecated since Python 3.8, use "async def" instead
```

E.g. we can use mock.AsyncMock (>= 3.8)

```python
cooper-mbp1:bandersnatch cooper$ grep -R asynctest *
dist/bandersnatch-4.0.3/setup.cfg:known_third_party = aiohttp,aiohttp_xmlrpc,asynctest,filelock,freezegun,packaging,pkg_resources,pytest,setuptools
requirements_test.txt:asynctest==0.13.0
setup.cfg:known_third_party = _pytest,aiohttp,aiohttp_socks,aiohttp_xmlrpc,asynctest,filelock,freezegun,keystoneauth1,mock_config,packaging,pkg_resources,pytest,setuptools,swiftclient
src/bandersnatch/tests/conftest.py:from asynctest import asynctest
src/bandersnatch/tests/conftest.py:    master.session = asynctest.MagicMock()
src/bandersnatch/tests/conftest.py:    master.session.get = asynctest.MagicMock(return_value=FakeAiohttpClient())
src/bandersnatch/tests/conftest.py:    master.session.request = asynctest.MagicMock(return_value=FakeAiohttpClient())
src/bandersnatch/tests/test_master.py:import asynctest
src/bandersnatch/tests/test_master.py:    master.rpc = asynctest.CoroutineMock(return_value=expected)  # type: ignore
src/bandersnatch/tests/test_master.py:    master.rpc = asynctest.CoroutineMock(return_value=[])  # type: ignore
src/bandersnatch/tests/test_master.py:    master.rpc = asynctest.CoroutineMock(return_value=None)  # type: ignore
src/bandersnatch/tests/test_master.py:    master.rpc = asynctest.CoroutineMock(  # type: ignore
src/bandersnatch/tests/test_master.py:    patcher = asynctest.patch("aiohttp.ClientSession", autospec=True)
src/bandersnatch/tests/test_package.py:import asynctest
src/bandersnatch/tests/test_package.py:    master.get_package_metadata = asynctest.CoroutineMock(  # type: ignore
[...]

PR #856, opened 2021-03-15
Replace asynctest with unittest.mock

ichard26:
Here's the PR I promised! 

This was easier than I had expected.

Resolves GH-787


cooperlees (approved):
Damn, was I this close. haha! I expected more.

cooperlees (changes requested):
Thanks! Exciting. Fix mypy plz.

cooperlees (approved):
Hells yes! Should be a lot less warnings now!

cooperlees:
The news check is broken. Adding tag.

PR #858, opened 2021-03-16
Remove debugging line that loads entire files into memory.

asrp:
Hi,

I saw RAM usage spike when bandersnatch starts up and found this line. It was effectively loading the entire file into memory as a string. Removing it seems to fix the memory issue.

Note: memory is used up even if log level is set above debug

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,7 @@
 
 - Unused storage plugins are loaded and cause non-fatal errors if dependencies are missing - `PR #799` - Thanks **electricworry**
 - Replaced usages of `asynctest` with `unittest.mock` in tests - `PR #807` and `PR #856` - Thanks **ichard26**
+- Remove debugging line that loads entire files into memory. - `PR #858` - Thanks **asrp**
 
 # 4.4.0 (2020-12-31)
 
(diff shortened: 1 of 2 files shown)


cooperlees (changes requested):
Awesome! Thanks for the fix. Always down to save memory.

Can you please just pop an entry into our bug fixes section in CHANGES.md for 5.0 with the #858 cited ...

asrp:
Done!

Feels weird thanking myself. :) Let me if it should be different.

ichard26:
> Feels weird thanking myself. :) 

Yeah it feels weird sometimes, but F/OSS work is often thankless for everyone involved. You deserve the thanks!

cooperlees (approved):
I feel you on the weird, e.g. I don’t tank myself here. But it’s nice to acknowledge people who see something and just fix it.

Thanks for having the monitoring and then debugging to find the simple fix.

QQ tho: do you run in debug mode? Would be nice if log / the strings didn’t interpolate unless debug is set. Don’t know how possible that is tho.

asrp:
I hadn't tested in debug mode.

You could potentially put it behind an if block and check `isEnabledFor`.

Issue #464, opened 2020-03-22
Remove terrible isinstance of Mock in mirror.py


cooperlees:
The way unittests are written today it causes us to mock the `synchronize()` method to allow for testing we're setting up mirror object correctly. Moving to asyncio has caused TypeErrors during the unittests so a very dirty isinstace was added:
```python
if not isinstance(mirror, Mock):  # type: ignore
    changed_packages = await mirror.synchronize()
```
It would be desirable to remove this. Options I see are:
- Rewrite unittests to not mock this the way we are today (need to maintain same coverage)
- Is there a better way to mock so we don't get this `TypeError`?

Other ideas?

PR #859, opened 2021-03-16
Remove terrible isinstance check in mirror.py

ichard26:
The way unittests were written caused us to mock the synchronize()
method to allow for testing we're setting up mirror object correctly.
Moving to asyncio has caused TypeErrors during the unittests so a very
dirty isinstace was added:

```python3
if not isinstance(mirror, Mock):  # type: ignore
    changed_packages = await mirror.synchronize()
```

This commit eliminates the isinstance check seen above by mocking the
synchronize() method directly with unittest.mock.AsyncMock while patching
the whole mirror class.

The reason why the isinstance check was needed in the first place is
that unittest.mock.patch replaces methods with unittest.mock.MagicMock
instances by default. Even when unittest.mock.AsyncMock is needed.

Resolves GH-464.


cooperlees (approved):
Absolutely gorgeous! Thanks for this!

I didn't have AsyncMock when I last attempted. This  does make it very nice and easy indeed.

PR #865, opened 2021-03-31
Point contributors to use Next PR Number

ichard26:
This should reduce the friction when adding a changelog entry by quite a bit.


cooperlees (approved):
Thanks!

Issue #864, opened 2021-03-24
Frequent timeout exceptions emitted from aiohttp


happyaron:
On our deployment we've observed frequent timeout exceptions emitted from aiohttp, with bandersnatch (4.4.0) + python (3.8.1):

In total 49 syncs:
1. `aiohttp.client_exceptions.ServerTimeoutError: Timeout on reading data from socket` (49 times)
2. `aiohttp.client_exceptions.ServerTimeoutError: Connection timeout to host <url>` (20 times)
    (where <host> are urls to https://files.pythonhosted.org)
This might be related to https://github.com/aio-libs/aiohttp/issues/4549

Also there is another exception from aiohttp which does not show a direct indication to timeout:
1. `aiohttp.client_exceptions.ClientPayloadError: Response payload is not completed` (25 times)
This might be related to https://github.com/aio-libs/aiohttp/issues/4581

Trying to use bandersnatch 3.6.0 (which uses asyncio w/ requests) + python (3.7.3), there's not any exceptions being raised.

Relevant parts of configuration look like this:
```
[mirrors]
json = true
timeout = 300
workers = 5
hash-index = false
stop-on-error = false
```

Our network has quite high latency than usual to reach to files.pythonhosted.org (~250ms rtt), while the latency is stable and packet loss is low). So we use a very long timeout setting in the configuration to make bandersnatch more patient to wait.

cooperlees:
Hi,

Thanks for the report. This is always a hard thing to tune to be right for all so we're going to have to play with aiohttp here.

1)
Have you tried with less workers? Or do you need that to max out throughput due to the high latency?

2)
Are you able to try our docker image (thus move to Python 3.9) and see if that helps things with aiohttp on your connection?

Does bandersnatch eventually battle away and get all the packages, or does it never due to the many timeouts/exceptions?
- Please feel free to fork bandersnatch and try different aiohttp setting. Interested to see what else we can tune. I don't have a bad connection so I've never sat down to play with this.

Another hack I've done on bad connection in past lives:
- Get a close VPS that's on a good connect and put a Forward Proxy there like squid

happyaron:
I tried the docker image with Python 3.9 with 20 syncs, the exceptions occur with similar rate, no real difference comparing to the Python 3.8 version.

[...]

PR #877, opened 2021-04-14
Run get_hash() in executor (#877)

happyaron:
Closes pypa/bandersnatch#864

Co-authored-by: @shankerwangmiao


cooperlees (approved):
Sounds good to me and I like the change as little as possible to help fix your problem.

We should (someone) eventually get to wrapping all I/O like this so we're truly non-blocking.

PR #883, opened 2021-04-19
storage: add move_file() and use in rewrite() and update_safe()

happyaron:
This commit avoids copying the file over on (local) filesystem in most circumstances which reduces disk IOs.

When `source` and `dest` are on the same filesystem `shutil.move()` makes use of `os.rename()` which is desired to be atomic, otherwise it will copy and delete like in our previous implementation.

### DIFF
--- a/src/bandersnatch_storage_plugins/filesystem.py
+++ b/src/bandersnatch_storage_plugins/filesystem.py
@@ -93,9 +93,7 @@ class FilesystemStorage(StoragePlugin):
         logger.debug(
             f"Writing temporary file {filepath_tmp} to target destination: {filepath!s}"
         )
-        self.copy_file(filepath_tmp, filepath)
-        logger.debug(f"Deleting temporary file: {filepath_tmp}")
-        self.delete_file(filepath_tmp)
+        self.move_file(filepath_tmp, filepath)
 
     @contextlib.contextmanager
     def update_safe(self, filename: PATH_TYPES, **kw: Any) -> Generator[IO, None, None]:
@@ -123,9 +121,7 @@ class FilesystemStorage(StoragePlugin):
             os.unlink(filename_tmp)
         else:
             logger.debug(f"Modifying destination: {filename!s} with: {filename_tmp}")
-            self.copy_file(filename_tmp, filename)
-            logger.debug(f"Deleting temporary file: {filename_tmp}")
-            self.delete_file(filename_tmp)
+            self.move_file(filename_tmp, filename)
 
     def compare_files(self, file1: PATH_TYPES, file2: PATH_TYPES) -> bool:
         """Compare two files, returning true if they are the same and False if not."""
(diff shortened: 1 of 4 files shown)


src/bandersnatch_storage_plugins/filesystem.py
```diff
@@ -138,6 +134,13 @@ def copy_file(self, source: PATH_TYPES, dest: PATH_TYPES) -> None:
         shutil.copy(source, dest)
         return
 
+    def move_file(self, source: PATH_TYPES, dest: PATH_TYPES) -> None:
+        """Move a file from **source** to **dest**"""
+        if not self.exists(source):
+            raise FileNotFoundError(source)
+        shutil.move(str(source), dest)
```
cooperlees, line 141: No biggy, but I think >= 3.8 move can handle PATH_TYPES ... (If wrong just ignore)

src/bandersnatch_storage_plugins/swift.py
```diff
@@ -720,6 +719,21 @@ def copy_file(
...
+        """Move a file from **source** to **dest**"""
+        if dest_container is None:
+            dest_container = self.default_container
+        dest = f"{dest_container}/{dest}"
+        with self.connection() as conn:
+            conn.copy_object(self.default_container, str(source), dest)
+            try:
+                conn.delete_object(self.default_container, str(source))
+            except swiftclient.exceptions.ClientException:
+                raise FileNotFoundError(str(source))
```
cooperlees, line 734: What's the advantage of raising `FileNotFoundError` rather than just letting the swift exception raise?

cooperlees (approved):
Looks good - Couple of nits - Wonder if there is any easy test we could add a test or two to stop our slippery slope down the coverage dropping ways ... (no biggy if not)

src/bandersnatch_storage_plugins/filesystem.py
happyaron: I tried on Python 3.9.2 and my tox run still shout at me about it so I added it...

src/bandersnatch_storage_plugins/swift.py
happyaron: This is just copied from SwiftStorage.delete_file(), I don't have an opinion on which is better, ;-)

src/bandersnatch_storage_plugins/filesystem.py
ichard26: Looking at the Python docs, this is expected behaviour:

> Changed in version 3.9: Accepts a path-like object for both src and dst.

https://docs.python.org/3/library/shutil.html#shutil.move

Our mypy configuration is configured to have mypy believe we are writing a Python 3.8 project.

https://github.com/pypa/bandersnatch/blob/164616ecb611714e9a0c410c8feb2ab86b0d5edd/mypy.ini#L2

cooperlees: Cool - thought it was 3.8 - my bad. This is all doing the correct things then! Thanks for checking all.

PR #887, opened 2021-04-24
Move to using asyncio.run

cooperlees:
- Now we're >= 3.8 we can just use asyncio's run() method to get into an async context

Tests:
- Make sure all tests pass
- Wrap some Master tests to ensure we have a running event loop for them to grab


cooperlees:
Damn coverage! Can never win.

PR #893, opened 2021-04-26
Update black from 20.8b1 to 21.4b0

ichard26:
Resolves #892 (i.e. supersedes)

I'm assuming this should get a skip news label


cooperlees (approved):
Thanks for the cleanup! I bet all those silly spaces in doc strings were me.

PR #894, opened 2021-04-26
Add a default ThreadPoolExecutor to storage plugin to wrap all IO

### LINKED ISSUE #584: Make filesystem storage plugin more asyncio friendly
Debugging a bug, I found that filesystem i/o is all serial and each function is synchronus and blocking on i/o.

Lets introduce a process / threadpool for asyncio to use to try up the i/o usage + eat more cpu.

Easy repro:
- Use whitelist to only download tensorflow (nice large wheels)
- Download 1/2 of tensorflow and cancel
- Restart, watch slow checksumming

I'm even using a slow USB hard disk on purpose too.

cooperlees:
- Have asyncio call all potential blocking IO in said ThreadPoolExectuor in heavy main workload - e.g. downloading many packages @ once
- Grepped popular blocking calls to find
- This should help stop starving the loop from CPU time
- Modified all modules that call storage block calls to wrap them in the sotrage class executor with max thread == workers

Testing:
- Fix unittests to pass
  - A lot just needed workers config
  - Some just needed to have a loop / become coroutines

Fixes #584

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,6 +17,7 @@
 - Migrated Markdown documentation from recommonmark to MyST-Parser + docs config clean up - `PR #879` - Thanks **ichard26**
 - Use `shutil.move()` for temp file management - `PR #883` - Thanks **happyaron**
 - Fixed logging bug in `SizeProjectMetadataFilter` to show it activated - `PR #889` - Thanks **cooperlees**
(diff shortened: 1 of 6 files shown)


cooperlees:
I never have people to review and CI looks good - So merging away! Hopefully less block the event loop in 5.0!

PR #920, opened 2021-05-21
Add autogenerated CLI help to docs index page.

### LINKED ISSUE #907: Automate `--help` Checking + Updating for Docs
It's nice to have `--help` in documentation. But, I always see this get out of date. I'd like to play one day with trying to get the docs CI to check if `--help` goes out of sync and if so, bonus points would be to auto update it somehow.

ichard26:
Using sphinx-argparse-cli instead of sphinxcontrib-programoutput because
I got convinced by gaborjbernat via Discord.

Unfortanately this requires to move parser configuration to its own
function because that's what sphinx-argparse-cli uses as input. This
isn't that painful tho.

---

Preview: https://bandersnatch.readthedocs.io/en/add-cli-help/

Resolves #907.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -62,7 +62,8 @@ instance/
 .scrapy
 
 # Sphinx documentation
-doc/html/
+docs/html/
+docs/_build/
 
 # PyBuilder
 target/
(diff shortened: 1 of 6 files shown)


cooperlees (approved):
Sweet. LGTM. Even gave us a coverage boost! haha

PR #922, opened 2021-05-22
Drop Travis CI and move Codecov

ichard26:
I'm using the codecov-action instead of the codecov Python package for
two reasons:

1. The action supports tokenless uploads
2. It's really confusing to have a local tox run try upload coverage

We do need to ask coverage to produce a XML report since Codecov doesn't
support coverage's custom sqlite(?) file.

Also, pull requests from an internal branch (i.e. one on pypa/bandersnatch
will be marked as a push coverage build by Coveralls since our
anti-duplicate-workflows system runs under the push even for such cases.


cooperlees (approved):
Cool - wonder if we get coverage per OS now … be interesting to compare

PR #928, opened 2021-05-28
Add support of downloading packages from alternative mirror 

### LINKED ISSUE #869: Support downloading files from alternative PyPI mirrors
bandersnatch currently downloads all files from the master PyPI archive, honoring the "url" in json respose which by default points to files.pythonhosted.org. Although creating a local master is theoretically possible, it is not practical when most people is mirroring the content with bandersnatch.

This feature is useful to sync most of the files from an existing, nearby mirror, for example when setting up a new server sitting next to an existing one for the purpose of load sharing. 

The design could be quite straightforward:
1. create a new configuration item which is disabled by default
2. if configured, try to download files from the specified alternative mirror first, and fallback to the "url" in json response from PyPI master
3. data integrity is checked by the sha256sum which bandersnatch always verifies for newly downloaded content

happyaron:
This PR adds two new configure options:

- `download-mirror`, default is `""`.
- `download-mirror-no-fallback`, default is `False`, always disabled when `download-mirror` is `""`.

`download-mirror` should set to the webroot of the mirror site, which is not always the root path of a domain like `https://pypi.pythonhosted.org`, but might look like `https://somemirror.example.com/pypi/` or other approperiate directory paths.

`download-mirror-no-fallback` is used mainly for testing/debugging purpose which forces not to fallback to use default URL, but could be useful for people who have network restriction implications.

Three test cases are added, together with the existing `test_mirror_sync_package`, we are covering:

- `test_mirror_sync_package`: `download-mirror` is not set, all original behavior, expect to succeed;
- `test_mirror_sync_package_download_mirror`: `download_mirror` is set to a working URL, and `download_mirror_no_fallback` is set to `True`, expect to succeed;
- `test_mirror_sync_package_download_mirror_fallback`: `download_mirror` is set to a not working URL, expect to fallback to default URL and succeed;
- `test_mirror_sync_package_download_mirror_fails`: `download_mirror` is set to a not working URL, and `download_mirror_no_fallback` is set to `True`, expect to raise an exception because download cannot continue.

Resolves #869.


src/bandersnatch/mirror.py
```diff
@@ -648,24 +652,49 @@ async def sync_release_files(self, package: Package) -> None:
...
+                        logger.info(
+                            "Continuing to next file after error downloading: " f"{url}"
+                        )
+                    else:
+                        logger.exception(
+                            "Continuing to next file after error downloading: " f"{url}"
+                        )
+                    # keep previous exception, also ignore non-default urls
+                    if not deferred_exception and len(download_urls) == (cnt + 1):
+                        deferred_exception = e
```
cooperlees, line 697: Follow up (or on this PR if you have cycles): this is complicated enough now maybe it warrants a function to explain in a doc string how we're looking up if we should use a mirror or fetch from URL in JSON (similar to what you put in the config).

src/bandersnatch/tests/test_mirror.py
```diff
@@ -440,6 +440,103 @@ async def test_mirror_sync_package_with_hash(
...
+    mirror.json_save = True
+    # Recall bootstrap so we have the json dirs
+    mirror._bootstrap()
+    # This download mirror URL does not work, forcing not to fallback
+    mirror.download_mirror = "https://not-working.example.com"
+    mirror.download_mirror_no_fallback = True
+
+    with mock.patch("bandersnatch.mirror.logger.exception") as mock_log:
+        await mirror.synchronize()
+        assert mock_log.call_count == 3
```
cooperlees, line 537: Maybe explain why it's 3

cooperlees (changes requested):
This is great! Thanks. Can we please just add some tests checking the config parsing + maybe move the new download url logic into a function.

Thanks for adding a useful feature to bandersnatch.

src/bandersnatch/tests/test_mirror.py
happyaron: 3 calls are:

- Tried and failed to get https://not-working.example.com/packages/any/f/foo/foo.zip
- Tried and failed to get https://not-working.example.com/packages/2.7/f/foo/foo.whl
- Declared error on syncing package foo@1 (mirror.py:L363)

src/bandersnatch/mirror.py
happyaron: Will give this a shot soon.

src/bandersnatch/tests/test_mirror.py
cooperlees: Thanks - Was more saying in comments so we can remember when we next look at this in a year or so :D - But this at least helps.

cooperlees: (As you did in https://github.com/pypa/bandersnatch/pull/928/commits/8df0b3bac45b62f5e5b465fc2a3c79ee5da5baf2 - Thanks)

cooperlees (approved):
Thanks! Much cleaner with that function. Hope this helps your needs and others!

That said, our poor test coverage is dropping tho. If you ever feel like helping bump that back up we'd love those contributions. I'll add an issue if I don't get around it it.

PR #931, opened 2021-06-01
Generate data-yanked tag in simple page

happyaron:
data-yanked is not generated in bandersnatch's simple page, this may confuse pip client because it will not install yanked release if not explictedly requested by user.


cooperlees (approved):
Many thinks for this! I didn't even know it was a thing.

cooperlees:
Seems we're hitting PyPI XMLRPC limits on some of the Integration Tests:
```
aiohttp_xmlrpc.exceptions.ApplicationError: HTTPTooManyRequests: The action could not be performed because there were too many requests by the client. Limit may reset in 1 seconds.
```

PR #932, opened 2021-06-01
Set up pre-commit.ci + update pre-commit config

ichard26:
post commit: it appears that pre-commit.ci only runs for pull requests, tags (or releases), and pushes to the default branch. I'm not sure whether this a dealbreaker. Personally it's nice to know during development that my branch isn't passing linting before submitting a PR but it's not strictly necessary.


ichard26:
Hmm, I guess we could document that `pre-commit.ci autofix` will make pre-commit.ci to trigger autofixing, but I'd guess it would just be annoying since it would do nothing for lint type warnings / errors (e.g. flake8, mypy, etc.).

.pre-commit-config.yaml
```diff
@@ -1,28 +1,41 @@
...
     hooks:
-    -   id: mypy
+      - id: mypy
         exclude: (docs/.*)
+
+  - repo: https://gitlab.com/pycqa/flake8
+    rev: 3.9.2
+    hooks:
+      - id: flake8
+        additional_dependencies: [flake8-bugbear]
```
cooperlees, line 38: Nice!

src/bandersnatch/tests/plugins/test_storage_plugins.py
```diff
@@ -670,7 +670,7 @@ def test_iter_dir(self) -> None:
 
         self.assertListEqual(
             list(sorted(base_path.iterdir(), key=lambda p: str(p))),
-            list(sorted([elem[0] for elem in lists], key=lambda p: str(p))),
+            list(sorted((elem[0] for elem in lists), key=lambda p: str(p))),
```
cooperlees, line 673: Just out of curiosity, what was the complaint here and from what tool? Sorting a list or tuple shouldn't differe here I thought :|

cooperlees (approved):
My only worry now with deleting `lint.yml` we don't see it break and it's a time bomb for developers to find.

I think we delete `lint` from tox config as well and change documentation to say for linking just run `pre-commit -a` ? Thoughts?

src/bandersnatch/tests/plugins/test_storage_plugins.py
ichard26: Blame pyupgrade, I didn't understand it either but also couldn't find a way to turn it off from a quick scan :/

.pre-commit-config.yaml
ichard26: Two birds with one stone :wink:

PR #933, opened 2021-06-01
Do some good ol' CI/CD housekeeping

ichard26:
- Move the doc_build workflow to latest Python
- Fix some step name issues
- Merge uvloop workflow into ci workflow because their only difference
  is whether the uvloop dependency is installed. Merging them will
  avoid slow and unnecessary dependency downloads and installs and will
  get rid of one more ugly skipped job.

I also added --use-feature=in-tree-build to our CI local installs to
be good users of sister project pip. Hopefully we don't encounter
any issues, but if we do then maybe we will be able to provide some
feedback. I only added it for CI because locally because 1) we
can't always trust the local pip version supporting the flag, and 2)
avoiding too much noise when things do break.


.github/workflows/ci.yml
```diff
@@ -53,5 +53,13 @@ jobs:
       env:
        TOXENV: INTEGRATION
       run: |
-        python -m pip install .
+        python -m pip install . --use-feature=in-tree-build
+        python test_runner.py
+
+    - name: Run Integration Test (uvloop)
+      if: matrix.os != 'windows-latest'
```
cooperlees, line 60: Nice - Only bummer here is it is now serially run after non uvloop ... But we can live with it

.github/workflows/doc_build.yml
```diff
@@ -3,23 +3,16 @@ name: doc_build
...
-        os: [ubuntu-latest]
-
+    runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2.3.4
 
-    - name: Set up Python ${{ matrix.python-version }}
+    - name: Set up latest Python
       uses: actions/setup-python@v2.2.2
-      with:
```
cooperlees, line 21: 🥰

cooperlees (approved):
Can live with the uvloop being run after non uvloop ...

cooperlees:
We could also use your parallel pytest speed wins here too ...

PR #941, opened 2021-06-09
data-yanked: fix exception when yanked-reason is present but empty

happyaron:
This PR fixes an exception when yanked-reason is present but empty, which is discovered when syncing the package borisml-0.0

Because it's a trivial fix to an exisiting commit, I don't think there should be a dedicated changelog entry...


cooperlees (approved):
Thanks for the fix. I guess this is worth of a 5.0.1 too ...

I would have thought if it existed it was set and if it was not set html.escapre would handle "" or None. Guess not.

cooperlees:
I'll add a changelog if I do a release just to show why we're releasing ...

PR #962, opened 2021-07-02
 Use update_safe when updating some repo-level metadata

### LINKED ISSUE #961: Use update_safe when updating some repo-level metadata
In the current implementation, when disk is full while wrapping up the sync, bandersnatch will truncate `last-modified`, `generation` and `status` files (and maybe more). This lead to a situation that the last successful sync's serial is lost, forcing next sync to start from verifying everything even the disk space issue is resolved.

Repacing some of the Path.write_text() and rewrite() to mitigate this problem.

happyaron:
In the current implementation, when disk is full while wrapping up the sync, bandersnatch will truncate last-modified, generation and status files (and maybe more). This lead to a situation that the last successful sync's serial is lost, forcing next sync to start from verifying everything even the disk space issue is resolved.

Repacing some of the Path.write_text() and rewrite() to mitigate this problem. This closes #961

### DIFF
--- a/src/bandersnatch/tests/plugins/test_storage_plugins.py
+++ b/src/bandersnatch/tests/plugins/test_storage_plugins.py
@@ -726,6 +726,12 @@ web{0}simple{0}index.html""".format(
 
     def test_find(self) -> None:
         base_path = self.mirror_base_path
+
+        # Clean up GitHub Actions environment on macOS tests
+        if sys.platform == "darwin":
+            env_garbage_path = os.path.join(self.mirror_base_path, "var")
+            shutil.rmtree(env_garbage_path, ignore_errors=True)
+
         self.assertEqual(self.base_find_contents, self.plugin.find(base_path))
 
     def test_open_file(self) -> None:
--- a/src/bandersnatch/tests/plugins/test_storage_plugins.py
+++ b/src/bandersnatch/tests/plugins/test_storage_plugins.py
@@ -726,8 +726,6 @@ web{0}simple{0}index.html""".format(
 
     def test_find(self) -> None:
         base_path = self.mirror_base_path
-        if self.backend == "swift":
-            base_path = base_path.lstrip("/")
         self.assertEqual(self.base_find_contents, self.plugin.find(base_path))
 
     def test_open_file(self) -> None:
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,7 @@
 
 - Add bandersnatch command line help to the documentation main page `PR #920` - Thanks **ichard26**
 - Generate data-yanked tag in simple page `PR #931` - Thanks **happyaron**
+- Protect repository metadata from being trashed when disk is full `PR #961` - Thanks **happyaron**
 
 # 5.0.0 (2021-4-28)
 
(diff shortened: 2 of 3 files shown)


cooperlees (approved):
Cool. Nice find and thanks for the fix!

cooperlees (changes requested):
Actually seems you've caused some different directories to appear with swift or the tests are flaky - I'm not sure what is going on here and don't have the time to debug right now.

happyaron:
I tried to re-run the tests on a macOS, with Python 3.8, and the tests are passing. See the attached log.
[tox.log](https://github.com/pypa/bandersnatch/files/6794670/tox.log)

cooperlees:
Wonder what's creating:
```
+ var
+ var/folders
+ var/folders/24
+ var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn
```

In the test, I can see we lstrip("/") :
https://github.com/pypa/bandersnatch/blob/master/src/bandersnatch/tests/plugins/test_storage_plugins.py#L729-L730

@ichard26 just caused tests to run ~10 minutes ago and they all passed, so I feel update_safe is causing something different with files on the filesystem here ... This is definitely a bug.

Non failed run on master: https://github.com/pypa/bandersnatch/actions/runs/1018394760

happyaron:
@cooperlees I'm trying to remove the lstrip() call in recent commit but those files still exist (with exactly the same path and name). It's confusing that normally things under `var/folders` on macOS are temp files/dirs like those ones under /tmp on Linux. They are expected to be created with some random name (at least not exactly the same across many rounds of tests).

cooperlees:
Is there a call / something we can get to get MacOS to place their temples elsewhere outside of our test env? I feel that would be good OR lets just add a `rmtree("path/to/test/dir/var")` before we run our asset if we're on MacOS with a comment about possible temp files existing in GitHub actions environment.

happyaron:
@cooperlees I take the second approach and it seems working. I can't reproduce the problem on my local macOS system (although Debian is my primary environment), and looking through our code I can't find where those files are expected to be created.

src/bandersnatch/tests/plugins/test_storage_plugins.py
```diff
@@ -726,8 +726,12 @@ def test_compare_files(self) -> None:
 
     def test_find(self) -> None:
         base_path = self.mirror_base_path
-        if self.backend == "swift":
-            base_path = base_path.lstrip("/")
+
+        # Clean up GitHub Actions environment on macOS tests
+        if sys.platform == "darwin":
+            env_garbage_path = os.path.join(self.mirror_base_path, "var")
+            shutil.rmtree(env_garbage_path, ignore_errors=True)
```
cooperlees, line 733: Thanks! This lets us know if more weirdness pops up we don't know about in the future.
- My only worry is we introduce this into peoples bandersantch directory structure, but if someone runs a mirror on MacOS they can add the fix somewhere for us - I think that's a low use case. Tests are mainly so us Mac Users can rely on the test suite passing while we develop.

cooperlees (commented):
Appreciate your persistence here!

cooperlees (approved):

PR #996, opened 2021-08-30
fix coverage badge link

LeoQuote:
the link of coverage badge is incorrect, should be bandersnatch


cooperlees (approved):
Sure, thanks. Copy paste fun here for sure.

PR #1007, opened 2021-09-09
Clean Up Deps + Update Unittest running Documentation

cooperlees:
- Move version to 5.1.0.dev0
- Remove version pinning from setup.cfg + add all deps + pin in requirements_s3.txt
- Add documentation on how to run S3 integration/unittests locally
- Change CI to use offical minio docker image

### DIFF
--- a/.github/workflows/ci_ubuntu.yml
+++ b/.github/workflows/ci_ubuntu.yml
@@ -18,6 +18,7 @@ jobs:
         image: minio/minio:latest
         ports:
           - 9000:9000
+        options: --entrypoint /usr/bin/docker-entrypoint.sh server /data
 
     strategy:
       matrix:
--- a/.github/workflows/ci_ubuntu.yml
+++ b/.github/workflows/ci_ubuntu.yml
@@ -15,10 +15,9 @@ jobs:
 
     services:
       minio:
-        image: minio/minio:latest
+        image: lazybit/minio:latest
         ports:
           - 9000:9000
-        options: --entrypoint '/usr/bin/docker-entrypoint.sh server /data'
 
     strategy:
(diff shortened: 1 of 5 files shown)


cooperlees:
I really don't get our coverage changes :| ...

.github/workflows/ci_ubuntu.yml
```diff
@@ -15,9 +15,10 @@ jobs:
 
     services:
       minio:
-        image: lazybit/minio:latest
+        image: minio/minio:latest
```
cooperlees, line 18: @LeoQuote - What was the main reason you used lazy bit/minio vs. the official image?

I'll see if I just need to pass more options. I'd like to use the official at it seems more maintained.

LeoQuote: the official minio image has an entrypoint `minio` while this command alone cannot start a minio server, and github action does not support adding argument or override entrypoint for image, I have to use the `lazybit/minio:latest`  image to get minio running.

cooperlees:
Bummer - Seems the services can't support custom arguments.

I guess options are:
1) Try add steps to start the container (it seems the ubuntu image has docker installed) manually
2) Ask minio to add a second entry point script that we could point at to start a no credential server that uses /data inside the container for GitHub Actions CI ...

LeoQuote:
There is an option that we download minio binary and setup a minio service , but a container can be more easy.maybe There is some kind of minio step in github action

cooperlees:
> There is an option that we download minio binary and setup a minio service , but a container can be more easy.maybe There is some kind of minio step in github action

Opened https://github.com/pypa/bandersnatch/issues/1008

PR #1043, opened 2021-09-25
Change pyup to run monthly

cooperlees:
- s3 deps release a lot so lets reduce churn here to be more manageable

### DIFF
--- a/.pyup.yml
+++ b/.pyup.yml
@@ -1,8 +1,21 @@
+assignees:
+  - cooperlees
+
+label_prs: "skip news"
+
+schedule: "every month"
+
 requirements:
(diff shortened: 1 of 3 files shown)


.pyup.yml
```diff
@@ -1,10 +1,23 @@
+assignees:
+  - cooperlees
+
+label_prs: "skip news"
```
ichard26, line 4: This is a nice touch :)

ichard26 (approved):
:rocket:

.pyup.yml
cooperlees: Less clicking is good.

PR #1066, opened 2021-12-10
Correctly log changes for synced packages.

### LINKED ISSUE #1065: Debug output for detailing package changes is wrong
The debug output that spits out the files changed for each package at the end of a mirror operation is broken.

Output should be something like `ipp added: [<list of files added for the ipp project>]`. Instead, the list is added to for each iteration, meaning that the list is every file for every project that has been spat out so far.

This can be a problem when performing an initial sync. With 344,479 projects to sync, and an average of 15 files for each, by iteration 100,000 each debug line is printing 1.5 million files. Even with debug off, the string generation still needs to happen and the process will be stuck seemingly hung for days.

forky2:
Fixes #1065


cooperlees (approved):
Ha! Nice logging bug fix. Many Thanks! Will try release once I get time to workout what is going on with our dependencies (other broken PRs)

PR #1073, opened 2021-12-22
Support Python 3.10 on S3

### LINKED ISSUE #1054: Support 3.10
- Seems we have issues with S3 plugin with 3.10

Need all CI to pass with 3.10 ... today it does not.

isidentical:
Fixes #1054

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,9 @@
+# Unreleased
+
+## Other
+
+- 3.10 is now supported `PR #1073` -- Thanks **isidentical**
+
 # 5.1.1 (2021-12-14)
 
 ## Bug fixes
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,7 +2,7 @@
 
 ## Other
 
-- 3.10 is now supported `PR #1073` -- Thanks **isidentical**
+- Python 3.10 is now supported `PR #1073` -- Thanks **isidentical**
 
 # 5.1.1 (2021-12-14)
 
--- a/requirements_s3.txt
+++ b/requirements_s3.txt
@@ -2,7 +2,7 @@ boto3==1.20.17
 botocore==1.23.17
 jmespath==0.10.0
 python-dateutil==2.8.2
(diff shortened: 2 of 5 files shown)


setup.cfg
```diff
@@ -86,7 +86,7 @@ uvloop =
     uvloop
 
 s3 =
-    s3path
+    s3path==0.3.3
```
cooperlees, line 89: I try not a version packages here so we don't lock people.
- Lets at least
```suggestion
    s3path>=0.3.3
```

cooperlees (changes requested):
Awesome. Thanks for this! Appreciate it.

Can we also add a 3.10 classifier to setup.cfg please to state we support 3.10 + fix the s3 extra dependency versioning and I'll merge.

setup.cfg
isidentical: Upps, my bad. I thought I did `>=`. Thanks!

requirements_s3.txt
```diff
@@ -2,7 +2,7 @@ boto3==1.20.17
 botocore==1.23.17
 jmespath==0.10.0
 python-dateutil==2.8.2
-s3path==0.3.2
+s3path>=0.3.3
```
cooperlees, line 5:
```suggestion
s3path==0.3.3
```
Here we pin to say this is the versions CI has tested.

cooperlees (approved):
Will fix before release.

CHANGES.md
```diff
@@ -1,3 +1,9 @@
+# Unreleased
+
+## Other
+
+- 3.10 is now supported `PR #1073` -- Thanks **isidentical**
```
ichard26, line 5:
```suggestion
- Python 3.10 is now supported `PR #1073` -- Thanks **isidentical**
```

ichard26 (approved):

ichard26:
Thank you for helping out @isidentical on so many projects in Cooper and I's world :slightly_smiling_face:

PR #1084, opened 2022-02-06
Move to PEP517 `build` package/frontend for packaging releases

cooperlees:
- `setup.py` direct invocations is deprecated
  - https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html
- Lets be a good **PyPA** package and move to build PEP517 buld frontend that should do the right things

Test

- Show Python version used for test:
```
crl-m1:ptr cooper$ python3 -V
Python 3.9.10
```
- `/tmp/test_build/bin/pip install --upgrade pip setuptools`
- `/tmp/test_build/bin/pip install --upgrade build pip setuptools twine`
- `/tmp/test_build/bin/python -m build`
  - Output: https://pastebin.com/rcmCp0cb

### DIFF
--- a/.github/workflows/pypi_upload.yml
+++ b/.github/workflows/pypi_upload.yml
@@ -15,13 +15,17 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v2.3.1
 
-    - name: Install latest pip, setuptools, twine, wheel
+    - name: Install latest build, twine
       run: |
-        python -m pip install --upgrade build pip setuptools twine wheel
+        python -m pip install --upgrade build twine
 
     - name: Build sdist & wheel
       run: |
         python -m build
+
+    - name: Run a twine check
+      run: |
+        twine --check dist/*
 
     - name: Upload to PyPI via Twine
       env:
--- a/.github/workflows/pypi_upload.yml
+++ b/.github/workflows/pypi_upload.yml
@@ -17,12 +17,11 @@ jobs:
 
     - name: Install latest pip, setuptools, twine, wheel
       run: |
-        python -m pip install --upgrade pip setuptools twine wheel
+        python -m pip install --upgrade build pip setuptools twine wheel
 
     - name: Build sdist & wheel
       run: |
-        python setup.py bdist_wheel
-        python setup.py sdist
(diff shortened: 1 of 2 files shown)


cooperlees:
Hi Gurus who I've asked for a review - Is there anything else bandersnatch should do here to be PEP517/8 compliant? Is there a need for a `pyproject.toml` to do anything else? We are all `setup.cfg` driven.

I'd like to get it all in tip top shape then go and see if we can move all other PyPA packages to use "what is the latest hotness" with packaging our modules ...

ichard26 (commented):
We will also want to declare bandersnatch's build requirements (i.e. setuptools) in `pyproject.toml` for good measure. Unfortunately I don't think we have a good lower bound written down anywhere to set for setuptools :(

cooperlees:
> We will also want to declare bandersnatch's build requirements (i.e. setuptools) in `pyproject.toml` for good measure. Unfortunately I don't think we have a good lower bound written down anywhere to set for setuptools :(

O ok, cheer. I'll copy black here and test to see how low I can go with setup tools and get a clean install. I just always test with latest and if it works move us to that version ... I'm a move forward kind of guy.

.github/workflows/pypi_upload.yml
```diff
@@ -17,12 +17,11 @@ jobs:
 
     - name: Install latest pip, setuptools, twine, wheel
       run: |
-        python -m pip install --upgrade pip setuptools twine wheel
+        python -m pip install --upgrade build pip setuptools twine wheel
```
pganssle, line 20: Do you still need setuptools and wheel here? Those should be pulled in in the isolated build environment, and if everything is working well the versions in the environment where `build ` is installed won't even be used.

cooperlees: Sweet, thanks. Done.

pyproject.toml
```diff
@@ -1,3 +1,3 @@
 [build-system]
-requires = ["setuptools>=42.0.0"]
+requires = ["setuptools>=43.0.0"]
```
cooperlees, line 2: Handy to know and this will be my min version moving forward.

PR #1086, opened 2022-02-12
Limit CI running to help PyPA

cooperlees:
- As request by di tune our jobs
  - To be fair our CI is pretty quick and we're not a busy project but I guess all helps
- Add only run on main branch push + pull requests to all on PR CI
- Also add the concurrency tweaks like `pip` to our main CI (`ci.yml` and `ci_ubuntu.yml`)
  - We test the S3 plugin in ubuntu and it has some extra deps to mimic the S3 API etc.
- Fix broken docker_upload branch name of "master" to "main"

Let me know if there are any more tweaks we can do to be a team plater.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,14 @@
 name: bandersnatch_ci
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
+  cancel-in-progress: true
 
 jobs:
   build:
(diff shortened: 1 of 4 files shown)


di (approved):
Thanks!

cooperlees:
The Mac, Windows + Ubuntu CI I would have expected to run here, no?

What "status to be reported" are the jobs waiting for?

di:
I think it's probably this conditional?

https://github.com/pypa/bandersnatch/blob/4ddf3e9d4c36dc173264301c240cfef924e51477/.github/workflows/ci.yml#L8-L12

cooperlees:
> I think it's probably this conditional?
> 
> https://github.com/pypa/bandersnatch/blob/4ddf3e9d4c36dc173264301c240cfef924e51477/.github/workflows/ci.yml#L8-L12

Thanks mate. Good catch.

Issue #1081, opened 2022-01-31
Remove swift support with Nvidia killing it?


cooperlees:
https://github.com/pypa/bandersnatch/runs/5011251999?check_suite_focus=true
- Need to workout new link

Also want to workout if swift is actually being use and if we want to keep the code around.
- It seems Nvidia has killed it after purchase?

cc: @techalchemy is it still worth keeping?

cooperlees:
https://github.com/pypa/bandersnatch/commit/082ac7b9f185993d17e8779f804c70ee190ba28a - Fixes docs for now.

I will remove swift support if no one speaks in a month or so.

cooperlees:
FWIW - I see the docker image still being pulled. So I will leave swift support until it causes pain. Then we can ask this question again.

cooperlees:
This is removed via #2081

PR #1088, opened 2022-02-13
s3: Fix typing + make functions same as Parent objects

### LINKED ISSUE #1074: Fix s3 Subclass of Storage missing Generators
Fixing typing for latest mypy I noticed some functions that are not same types as parent
- The functions that need to be fixed are commentated in https://github.com/pypa/bandersnatch/pull/1071

cooperlees:
- Some functions were not made Generators that should be
- Also file filesystem `open_file` to not need a dict + copy to s3 child object
- Remove `tf.has_changed = False` from filesystem + s3 update_safe fucntion as it's not used

Ignoring swift plugin due to #1081

Fixes #1074

### DIFF
--- a/src/bandersnatch_storage_plugins/filesystem.py
+++ b/src/bandersnatch_storage_plugins/filesystem.py
@@ -7,7 +7,7 @@ import os
 import pathlib
 import shutil
 import tempfile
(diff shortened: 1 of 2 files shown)


cooperlees:
Massive coverage win! lol

PR #1110, opened 2022-04-11
Enhanced platform filter to filter out specific Python minor versions.

davidkim83:
- An attempt to filter out Python versions when repository maintainers do not need specific (ex. old, unsupported) versions of Python.
- Modifications made to `src/bandersnatch_filter_plugins/filename_name.py` 
- Mirror filtering docs & test case modified to cover for the modifications


src/bandersnatch/tests/plugins/test_filename.py
```diff
@@ -159,4 +189,4 @@ def test_exclude_platform(self) -> None:
         assert sum(f["flag"] == "DROP" for r in rv for f in r) == 0
 
         # the release "0.2" should have been deleted since there is no more file in it
-        assert len(pkg.releases.keys()) == 2
+        assert len(pkg.releases.keys()) == 3
```
cooperlees, line 192: Why is this 3 now? Maybe add that to the comment.

src/bandersnatch/utils.py
```diff
@@ -139,3 +139,37 @@ def bandersnatch_safe_name(name: str) -> str:
...
+    pep425_pyver_str_pp = f"-pp{_version_without_dot}-"
+    pep425_pyver_str_py1 = f"-py{_version_with_dot}-"
+    pep425_pyver_str_py2 = f"-py{_version_with_dot}."
+
+    _versions.extend([pep425_pyver_str_cp])
+    _versions.extend([pep425_pyver_str_ip])
+    _versions.extend([pep425_pyver_str_jy])
+    _versions.extend([pep425_pyver_str_pp])
+    _versions.extend([pep425_pyver_str_py1])
+    _versions.extend([pep425_pyver_str_py2])
```
cooperlees, line 173: Does this logic exist in the Python packaging or another core library? Just want to avoid maintaining it if it's in a core library.
- My quick look at [packaging](https://github.com/pypa/packaging) says no ...

src/bandersnatch_filter_plugins/filename_name.py
```diff
@@ -16,6 +17,24 @@ class ExcludePlatformFilter(FilterReleaseFilePlugin):
...
+        "py3.3",
+        "py3.4",
+        "py3.5",
+        "py3.6",
+        "py3.7",
+        "py3.8",
+        "py3.9",
+        "py3.10",
+    ]
+
```
cooperlees, line 37: Wonder if there is any way we can dynamically generate this when new versions come out :|
- Probably not worth the effort or complexity ..

(I just like to try and avoid things that need to be updated every new python release as it becomes operational overhead)

This is outside the scope of this PR, I'll make another issue for it when we merge, unless someone reading comes up with a clever idea.

cooperlees (changes requested):
Many thanks for contributing. Looks great. Lets see how CI goes and lets maybe add some comments / answer my questions here just to be sure.

src/bandersnatch/utils.py
ichard26: There is a dictionary of interpreters and their short names which would allow to not hard code a list at least: https://packaging.pypa.io/en/latest/tags.html#packaging.tags.INTERPRETER_SHORT_NAMES

src/bandersnatch/tests/plugins/test_filename.py
davidkim83: I added another 'release' in the test data - lines 150 ~ 176 - that shouldn't be filtered out.
The test still checks the count of 'KEEP's and 'DROP's, which I think is sufficient for the new filtering logic.

src/bandersnatch_filter_plugins/filename_name.py
davidkim83: Here's a sample of a way to generate new versions.
Someone would still have to increment the major/minor versions, but its better than having to increment the list by hundreds (e.g. 3.98, 3.99) if it ever comes to that.
This assumes that Python never changes their version scheming too.

```python
_versions: List[str] = []
_ver_separator = '.'

_py_mjr_ver_min = 2
_py_mjr_ver_max = 3
_py_mjr_vers = list(map(str, list(range(_py_mjr_ver_min, _py_mjr_ver_max+1))))

_py2_mnr_ver_min = 4
_py2_mnr_ver_max = 7

_py3_mnr_ver_min = 0
_py3_mnr_ver_max = 10

_py2_mnr_vers = list(map(str, list(range(_py2_mnr_ver_min, _py2_mnr_ver_max+1))))
_py3_mnr_vers = list(map(str, list(range(_py3_mnr_ver_min, _py3_mnr_ver_max+1))))

_conf_prefix = 'py'

for mjr in _py_mjr_vers:
    if "2" == mjr:
        _mnr_versions = [_conf_prefix + mjr + _ver_separator + mnv_ver for mnv_ver in _py2_mnr_vers]
    elif "3" == mjr:
        _mnr_versions = [_conf_prefix + mjr + _ver_separator + mnv_ver for mnv_ver in _py3_mnr_vers]
    
    _versions.extend(_mnr_versions)


print(_versions)
# ['py2.4', 'py2.5', 'py2.6', 'py2.7', 'py3.0', 'py3.1', 'py3.2', 'py3.3', 'py3.4', 'py3.5', 'py3.6', 'py3.7', 'py3.8', 'py3.9', 'py3.10']

```

src/bandersnatch/utils.py
davidkim83: Here's my attempt at incorporating the comments - would you check if this would suffice?
I found that two types of separators are needed for an edge case: [Numpy 1.7.1](https://pypi.org/project/numpy/1.7.1/#files)

```python
from packaging.tags import INTERPRETER_SHORT_NAMES
from typing import Dict, List


def parse_version(version: str) -> List[str]:
    _versions: List[str] = []
    
    interpreters = INTERPRETER_SHORT_NAMES.values()
    tag_separator1 = '-'
    tag_separator2 = '.'
    
    _versions.extend([tag_separator1 + i + version + tag_separator1 for i in interpreters])
    _versions.extend([tag_separator1 + 'py' + version + tag_separator2])
    
    return _versions


parse_version('3.7')
# ['-py3.7-', '-cp3.7-', '-pp3.7-', '-ip3.7-', '-jy3.7-', '-py3.7.']

```

image

cooperlees (changes requested):
Looks good. lets fix the `parse_version` and this feels ready to merge. Many thanks for this.

- Docs building failing I'm trying to fix in https://github.com/pypa/bandersnatch/pull/1112 but can't workout the GitHub links :|

davidkim83:
Thanks for the review so far - I personally learned a lot from this process of contribution.
If there's anything else I should do just let me know.

FYI, I did a live test by cloning Numpy package only, where the repo size reduced:
- from **26 GB** in full
- to **8.5 GB** by __filtering all but 3.6 and 3.8__ (it still gets source codes and generic py3, py2.py3 wheels)
- to **4.6 GB** by __filtering windows/macos binaries__ in addition to the python versions

Since repo size/cost & clonging time is directly proportional to the number & size of files, this is sure to improve repo cloning for many admins behind firewalled on-prem environments.

cooperlees (approved):
All looks good and thanks for taking feedback and working on this!

Will try get a release out in the next week or so.
- Need to find time to fix the docs first

PR #1113, opened 2022-04-18
Doc updates for Swift

tipabu:
Note that actually _using_ Swift as a backend seems to be currently busted -- I've been hacking on it to get it working, though.


cooperlees (approved):
Awesome. As you can tell, I've never used swift ... So I appreciate the PR here a lot. @techalchemy is the author of all this, but have not heard from him in a long time.

PR #1115, opened 2022-04-30
Remove deprecated --use-feature=in-tree-build in CI

ichard26:
pip now always does in-tree builds, see also https://github.com/pypa/pip/pull/11001


cooperlees (approved):
Thanks - I don't even know why we did it ...

PR #1116, opened 2022-04-30
Add pip-audit to pre-commit

cooperlees:
- Add checking our deps for vulnerabilities to CI

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -46,3 +46,5 @@ repos:
 
 ci:
   autofix_prs: false
+  # Leave pip-audit to only run locally and not in CI
+  skip: [pip-audit]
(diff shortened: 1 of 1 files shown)


cooperlees:
https://github.com/pre-commit/pre-commit/issues/2368 - Asking if this can be fixed or if I just move running pre-commit back to a GitHub action.

ichard26:
I echo Anthony's sentiments how this is better as a regular GHA job. Internet performance of contributors is not guaranteed and while not many people are contributing here anyway, that just means pip-audit won't be run often or regularly. IMO it would be better put this in its own workflow that runs daily and on changes to the requirement files.

cooperlees:
I can never win with pre-commit. I just went to use it cause it seems what the open source world uses. I dislike it in general and would rather move everything out of pre-commit to be honest.

PR #1120, opened 2022-05-01
CI Tweaks + Fixes + Doc Update

cooperlees:
- Contributing Docs updated
- Remove "lint" from tox.ini
  - We were running py3 twice due to this - This should 1/2 "Run Unittests" step time for each OS
- Make tox run `pytest` with `--strict-markers`
- Set asyncio strict mode in pytest.ini as it's going to be default soon
  - I don't know how to fix the asyncio warnings tho - e.g. getting running loop in `__init__`

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -18,6 +18,8 @@ Please make sure you system has the following:
 
 - Python 3.8.0 or greater
 - git client
+- docker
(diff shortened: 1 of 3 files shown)


cooperlees:
Ok, cause of using `test_runner.py` and tox supplying py3 we were not running twice. I just did locally cause I ran `tox` directly with no environment set (via ENV Var or cli)

PR #1122, opened 2022-05-02
Add PEP629 Version to Simple API HTML

### LINKED ISSUE #1121: Add PEP629 Simple API Version into generated HTML
PEP 629 1.0's PEP 503's version of the simple API. Lets add that into the files we generate to uphold the standard.

Today it's just adding:
```html
  <meta name="pypi:repository-version" content="1.0">
```
- In the <head> section of the simple HTML

cooperlees:
- Add 1.0 to our HTML to indicate we only support that
- Make a static class variable for version number so it will be easy to update in our two write locations
- Update test expectations to see it work how we want

Fixes #1121

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 ## Other
 
+- Add PEP 620 header to Simple API HTML `PR #X`
 - Add S3 Docker Image building `PR #1092`
 - Move Docker containers to Python 3.10 `PR #1092`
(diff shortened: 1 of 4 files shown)


brettcannon (commented):
I can't actually provide an explicit review state, but LGTM!

cooperlees:
> I can't actually provide an explicit review state, but LGTM!

Guessed as much. Already released.

Issue #1134, opened 2022-06-13
Update to GitHub action actions/setup-python@v4 broke docs build


cooperlees:
Lets debug and fix. 
- Example Fail: https://github.com/pypa/bandersnatch/runs/6871377009?check_suite_focus=true
- PR introducing fail: https://github.com/pypa/bandersnatch/pull/1132
  - I think it was missed as we don't always run doc_build unless certain files are changed ... Maybe we should add the docs yaml config too

cooperlees:
Seems this has also broke our dependency vulnerability check: https://github.com/pypa/bandersnatch/runs/6874663934?check_suite_focus=true
- I think I'll roll back today until we workout what the issue is

ichard26:
The problem is that version 4 doesn't have a default Python version. You have to specify the version either inline in the YAML or in a file.

PR #1135, opened 2022-06-14
Specify Python version in CI

ichard26:
actions/setup-python starting with version 4 doesn't have a default Python version. Also remove useless conditional in doc_build.

Fixes GH-1134.

### DIFF
--- a/.github/workflows/doc_build.yml
+++ b/.github/workflows/doc_build.yml
@@ -9,14 +9,14 @@ on:
 jobs:
   lint:
     name: html + linkcheck build
-    if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
-
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
 
     - name: Set up latest Python
       uses: actions/setup-python@v4
+      with:
+        python-version: "*"
 
     - name: Install latest pip, setuptools + tox
       run: |
--- a/.github/workflows/pypi_upload.yml
+++ b/.github/workflows/pypi_upload.yml
@@ -14,6 +14,8 @@ jobs:
 
     - name: Set up Python
       uses: actions/setup-python@v4
+      with:
+        python-version: "*"
 
     - name: Install latest build, twine
       run: |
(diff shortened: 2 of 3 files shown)


.github/workflows/doc_build.yml
```diff
@@ -9,14 +9,14 @@ on:
...
-    if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
-
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
 
     - name: Set up latest Python
       uses: actions/setup-python@v4
+      with:
+        python-version: "*"
```
cooperlees, line 19: Does * mean latest available?

hugovk: I don't find `*` documented at https://github.com/actions/setup-python. They have:

```yml
python-version: '3.x' # Version range or exact version of a Python version to use, using SemVer's version range syntax
```

For the Black PR, I saw one of the [other workflows](https://github.com/psf/black/blob/6c1bd08f16b636de38b92aeb2e0a1e8ebef0a0b1/.github/workflows/upload_binary.yml#L39) already had `*` so I copied and pasted from there!

That was added in https://github.com/psf/black/pull/2260 in commit [psf/black@`2fcc183` (#2260)](https://github.com/psf/black/pull/2260/commits/2fcc183a07e703c8d6166a7bf533a463a885beb9), and there's no docs reference. Maybe @felix-hilden remembers where it came from? Lucky guess?

cooperlees: Stuff works so we can merge and make better if we learn :D Thanks!

cooperlees (approved):

.github/workflows/doc_build.yml
felix-hilden: The [doc](https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-python#using-a-specific-python-version) is linked in the PR! But the gist is: it's a SemVer range, so indeed the latest. `3.x` would not allow Python 4 versions, which obviously is a *huge* issue 😉

PR #1142, opened 2022-07-06
Use storage backend to verify and delete package

LeoQuote:
This pr is about using more storage backend when deleting or verifying packages.

* using path.unlink() to delete packages
* use writeBytes() and readBytes() to copy file instead of `shutils.move`

fix https://github.com/pypa/bandersnatch/issues/1137

## proposol

symlink should not be used in this project, we should not rely on it since s3 does not support that. 

symlink should be removed , move to other method , or using copy instead.


src/bandersnatch/mirror.py
```diff
@@ -654,9 +653,7 @@ def save_json_metadata(self, package_info: Dict, name: str) -> bool:
         # In 4.0 we move to normalized name only so want to overwrite older symlinks
         if self.json_pypi_symlink(name).exists():
             self.json_pypi_symlink(name).unlink()
-        self.json_pypi_symlink(name).symlink_to(
-            os.path.relpath(self.json_file(name), self.json_pypi_symlink(name).parent)
-        )
+        self.json_pypi_symlink(name).write_bytes(self.json_file(name).read_bytes())
```
LeoQuote, line 656: In this case , symlink is replaced with `write_bytes` and `read_bytes`, these are identical to copy files

src/bandersnatch/delete.py
```diff
@@ -45,11 +45,30 @@ async def delete_path(blob_path: Path, dry_run: bool = False) -> int:
...
+    simple_base_path: Path, package: str, dry_run: bool = True
+) -> None:
+    if dry_run:
+        logger.info(f"[dry run]rm simple page of {package}")
+        return
+    simple_dir = simple_base_path / package
+    simple_index = simple_dir / "index.html"
+    simple_index.unlink(missing_ok=True)
+    if simple_dir.exists():
+        simple_dir.rmdir()
```
cooperlees, line 58: I think this needs to be a little smarter if [hash-index](https://github.com/pypa/bandersnatch/blob/main/src/bandersnatch/default.conf#L50) is set.

- We should check if the hash is empty and delete if so maybe - Haven't looked into if non filesystem cares here or not ...

src/bandersnatch/delete.py
```diff
@@ -75,7 +94,12 @@ async def delete_packages(config: ConfigParser, args: Namespace, master: Master)
             logger.error(f"{json_full_path} does not exist. Pulling from PyPI")
             await get_latest_json(master, json_full_path, config, executor, False)
             if not json_full_path.exists():
-                logger.error(f"Unable to HTTP get JSON for {json_full_path}")
+                logger.error(
+                    f"Unable to HTTP get JSON for {json_full_path}, "
+                    f"blob files will not be cleaned."
+                )
+                await delete_simple_page(simple_path, canon_name, dry_run=args.dry_run)
+                await delete_simple_page(simple_path, package, dry_run=args.dry_run)
```
cooperlees, line 102: So, what's the logic here - I'm missing some context:
- If no JSON exists, we should delete the simple pages for a cleanup?

src/bandersnatch/main.py
```diff
@@ -156,7 +162,9 @@ async def async_main(args: argparse.Namespace, config: ConfigParser) -> int:
     elif args.op.lower() == "verify":
         return await bandersnatch.verify.metadata_verify(config, args)
     elif args.op.lower() == "sync":
-        return await bandersnatch.mirror.mirror(config, args.packages)
+        return await bandersnatch.mirror.mirror(
+            config, args.packages, not args.skip_root
```
cooperlees, line 166: Lets do this in a dedicated PR - Seems scope creep for this PR title.

What's your use case here so I can just understand.

src/bandersnatch/utils.py
```diff
@@ -118,6 +116,15 @@ def recursive_find_files(files: Set[Path], base_dir: Path) -> None:
         recursive_find_files(files, directory)
 
 
+def find_all_files(files: Set[Path], base_dir: Path) -> None:
+    for f in base_dir.rglob("*"):
+        if not f.is_file():
+            continue
+        if hasattr(f, "keep_file") and f.name == f.keep_file:  # type: ignore
+            continue
+        files.add(f)
```
cooperlees, line 125: Thanks for the move. Lets ensure we get a unittest around here - I'm a big fan of util functions always having good unittest coverage as they should be able to (and this can be) easily tested :)

cooperlees (commented):
Looks good. Lets maybe try split up into some smaller PRs and get some unittest coverage on the smaller PRs ... Lots of smaller changes easier to:
- Avoid bugs
- Test 
- Review

Thanks for the efforts here.

LeoQuote:
This is a draft so I skipped tests, I was waiting for your review, thanks for your review!
I’ve noticed the json file is placed in two places ( pypi and json ) , is there some protocol that we should follow? Can we reduce it to one single place?

src/bandersnatch/main.py
LeoQuote: Yes, will do

src/bandersnatch/delete.py
LeoQuote: I encountered some cases that I didn’t enable json_save and I wanted to clean a package that was deleted in pypi.org , the delete command would raise an error since json file was not present and cannot be get in pypi.org .

This is some kind of quick fix for this case( just delete the simple page, ignore packages completely), my initial thought was that bandersnatch verify would take care of those packages that does not belong to any package. But I later came to another solution: I can “generate” package info from simple page instead of json page.

How do you think we should we handle this case?

cooperlees: I feel here we should know (by the config) that we don't save JSON and either pull from the internet or we can just cleanup I guess. I'm not sure I 100% follow this flow.

cooperlees:
Rightio then.

The `JSON API` isn't standardized in a PEP, but is super handy so I started mirroring it years ago. I dislike where it's URL route is located (i.e. `pypi/PKG_NAME/json` - e.g https://pypi.org/pypi/bandersnatch/json) as if you had multiple PyPI mirrors internally you can't have the JSON API due to the `/pypi` base route ...

So `pip` etc. today will look at the `pypi/PKG_NAME/json` endpoint/route - So we could just get away with that. But I was trying for a few years to change this URL, with little success.

LeoQuote:
OK, In that case I will keep it that way.

LeoQuote:
The pre-commit.ci's behavior is quite confusing, the result is different in the ci check and the local check, my local check would say `# type: ignore` is unused, while the ci check says the type is not matched.... I dont know what to do to fix it

src/bandersnatch/verify.py
```diff
@@ -60,11 +63,13 @@ async def get_latest_json(
         if e.status == 404:
             # A 404 means that the package has been removed from PyPI.
             # Allow function to continue, and remove package files if applicable.
-            pass
+            # write a blank json file to make the deletion process go through
+            write_fake_json(json_path)
```
LeoQuote, line 67: This is a bit hack, I wrote a fake json in json_path to make "clean blob" process go through, please let me know if this is not a good practice

CHANGES.md
```diff
@@ -6,6 +6,7 @@
 - Added support to exclude specific Python minor versions `PR #1110` - Thanks **davidkim83**
 - Add S3 Docker Image building `PR #1092`
 - Add `--skip-simple-root` option for `bandersnatch sync` command `PR #1145`
+- Add delete and verify support in s3 storage backend `PR #1142`
```
cooperlees, line 9: 5.2.0 is realeased :) Make a new # X.X.X (2022-MM-DD) section please :D

src/bandersnatch/mirror.py
```diff
@@ -626,10 +626,10 @@ def _save(self) -> None:
     """
 
     def json_file(self, package_name: str) -> Path:
-        return Path(self.webdir / "json" / package_name)
+        return self.webdir / "json" / package_name
 
     def json_pypi_symlink(self, package_name: str) -> Path:
-        return Path(self.webdir / "pypi" / package_name / "json")
+        return self.webdir / "pypi" / package_name / "json"
```
cooperlees, line 632: Thanks - I believe this was done to make earlier version of mypy happy.

src/bandersnatch/storage.py
```diff
@@ -340,7 +340,7 @@ def load_storage_plugins(
         config = BandersnatchConfig().config
     if not enabled_plugin:
         try:
-            enabled_plugin = config["mirror"]["storage-backend"]
+            enabled_plugin = config.get("mirror", "storage-backend")
```
cooperlees, line 343: Why is this needed? Now we'll never hit to KeyError we catch to default to filesystem ...

src/bandersnatch/tests/test_verify.py
```diff
@@ -196,12 +207,11 @@ async def test_delete_unowned_files() -> None:
...
 @pytest.mark.asyncio
 async def test_get_latest_json(monkeypatch: MonkeyPatch) -> None:
-    config = FakeConfig()
     executor = ThreadPoolExecutor(max_workers=2)
     json_path = Path(gettempdir()) / f"unittest_{os.getpid()}.json"
     master = Master("https://unittest.org")
-    master.url_fetch = do_nothing  # type: ignore
-    await get_latest_json(master, json_path, config, executor)  # type: ignore
+    master.url_fetch = fake_fetch  # type: ignore
+    await get_latest_json(master, json_path, executor)
```
cooperlees, line 214: Weird - We don't assert anything here :| - Should we check a json file exists or something?

src/bandersnatch/tests/test_verify.py
```diff
@@ -238,9 +248,11 @@ class FakeArgs:
     jsonpath.mkdir(parents=True)
     jsonfile = jsonpath / "bandersnatch"
     jsonfile.touch()
-    all_package_files: List[str] = []
+    all_package_files: List[Path] = []
```
cooperlees, line 251: Sweet - Always good to get away from strs for paths.

src/bandersnatch/utils.py
```diff
@@ -212,3 +213,9 @@ def parse_version(version: str) -> List[str]:
     )
 
     return _versions
+
+
+def write_fake_json(json_path: Path) -> None:
+    if not json_path.parent.exists():
+        json_path.parent.mkdir(parents=True)
+    json_path.write_text(json.dumps({"releases": {}}))
```
cooperlees, line 221:
```suggestion
    json_path.write_text(json.dumps({}))
```
Might as well make it as small as possible ...

src/bandersnatch/verify.py
cooperlees: So you need this just so other deletion code does not error? Can we just catch an exception if the json file does not exist and skip writing out this fake JSON file all together?

src/bandersnatch/verify.py
```diff
@@ -131,14 +134,19 @@ async def verify(
...
             except Exception as e:
                 on_error(stop_on_error, e, package=json_file)
         else:
             logger.info(f"[DRY RUN] Would of grabbed latest json for {json_file}")
 
+    if (
+        hasattr(json_full_path, "keep_file")
+        and json_full_path.name == json_full_path.keep_file  # type: ignore
+    ):
+        logger.debug(f"Skipping keep file {str(json_full_path)}")
```
cooperlees, line 147:
```suggestion
        logger.debug(f"Skipping deleting JSON file due to keep_file attribute being set {str(json_full_path)}")
```

src/bandersnatch/verify.py
```diff
@@ -241,7 +254,7 @@ async def metadata_verify(config: ConfigParser, args: Namespace) -> int:
 
     logger.info(f"Starting verify for {mirror_base_path} with {workers} workers")
     try:
-        json_files = await loop.run_in_executor(executor, os.listdir, json_base)
+        json_files = list(x.name for x in json_base.iterdir())
```
cooperlees, line 257: Can't this block the async event loop?

cooperlees (changes requested):
Looks good. Thanks for adding this - Lets just fix the few things:
- CHANGELOG

And I just want to understand why we need the fake JSON + some other questions and suggestions.

src/bandersnatch/storage.py
LeoQuote: This is mainly for the test, the mocked config in tests only support `.get` method

src/bandersnatch/tests/test_verify.py
LeoQuote: I'll look into that, I didnt notice that as I was just doing some refactoring , assuming this is just fine.

src/bandersnatch/verify.py
LeoQuote: Certainly, the only downside is that the nested `if` and `try` make the indent comparely long, I'll try that.

src/bandersnatch/verify.py
LeoQuote: I'm not so familliar with this, will try to do that!

LeoQuote: I didnt find a way to achive that , listdir() yield a `generator` like `range` , If I have to do this in another executor , I have to wrap this up and call it with executor, which seems to be unneccessary ? What do you think?

src/bandersnatch/tests/test_verify.py
LeoQuote: A tiny assert is added, and it passed! This test if for `get_latest_json` function, which is to fetch json from pypi.org, this test patched url_fetch method and return a fake text , this test assert the fake text is written to json file

src/bandersnatch/storage.py
cooperlees: That is not a desirable thing to do when you break the actual codes logic just to make a mock to work.

What ever test is mocking here, can we just change it to read in once of the test config files into a ConfigParser object and change this back please.

cooperlees: Actually you could just get rid of the `try/except` and make the .get default to "filesystem" right to get same behavior ...

cooperlees (changes requested):
Please rebase the PR with main (master) + fix that config parser bug you've introduced. Making the .get() default to "filesystem" is prob best.

Do you have time in the next day or so to do this or should I? I'd like to merge this before my big https://github.com/pypa/bandersnatch/pull/1154 as one of us is going to have to ensure large merge conflicts possibly.

LeoQuote:
Thanks for your review! I'll make this change ASAP, I think this tiny change could be done before Saturday.

cooperlees (approved):
Lets land this ...

LeoQuote:
Please wait for a sec as I found something wrong in changelog and storage calls

PR #1145, opened 2022-07-11
make simple root page update optional

LeoQuote:
Genrating simple root page is sometimes time-consuming , and it is not really used by pip( pip will use `https://pypi.org/simple/my-package/`  directly  ), thus not so important, this option could lead to an unsynced simple root page , but no critical damage is made.


cooperlees (changes requested):
Awesome stuff with a great unit test. Lets just make the argument name a little more explicit so its harder to mis understand.

cooperlees (approved):
Thanks for your contribution! This could be useful for peoples.

PR #1161, opened 2022-07-31
Add PEP 691 config to banderx

cooperlees:
banderx is an example way to serve a filesystem bandersnatch.
This PR adds an example to respect PEP691 clients.

- Updated README.md
- Added a README.md to banderx dir to explain how to build + start

Test: Ran the container and ensured nginx did not error and returned via `curl`

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,9 +6,10 @@
 
 ----
 
-This is a PyPI mirror client according to `PEP 381` + `PEP 503`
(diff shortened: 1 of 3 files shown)


cooperlees:
Really need to fix this bot or remove it. It's annoying always getting this email for diffs that don't change any code :(

PR #1227, opened 2022-09-27
Fix missing method call

JelleZijlstra:
Noticed this in https://github.com/python/mypy/pull/13686#issuecomment-1251019429


cooperlees (approved):
Thanks. Would this have always been false? I wonder how my tests did not hit this …

… or our own mypy runs …

JelleZijlstra:
It would always be true because method objects are truthy. I didn't try to figure out if that manifests as a user-visible bug.

Your mypy runs didn't catch it because the check was off by default :)

PR #1232, opened 2022-10-01
Added possibility to limit hours range of mirroring activity

mbacicc:
Added a runner optional argument, which does not break backwards compatibilty to limit the working hours of the mirroring process. This could be useful for peolpe who want to have full download bandwidth for their work / activities during the day and sync the repo during the night.

Hour range is expressed as `10-16`, `20-6` (with day-carry) or single hour `3`.

The optional parameter name is `--hours-range` and it defaults to `0-23` for backward compatibility.

### DIFF
--- a/src/runner.py
+++ b/src/runner.py
@@ -8,13 +8,14 @@ import sys
 from datetime import datetime
 from subprocess import CalledProcessError, run
 from time import sleep, time
+from typing import List
 
 
-def parseNumList(string):
-    m = re.match(r"(\d+)(?:-(\d+))?$", string)
+def parseHourList(time_str: str) -> List[int]:
+    m = re.match(r"(\d+)(?:-(\d+))?$", time_str)
     if not m:
         raise argparse.ArgumentTypeError(
-            f"'{string}' is not a range. Expected '0-5', '20-6' or '1'."
+            f"'{time_str}' is not a range. Expected '0-5', '20-6' or '1'."
         )
     start = int(m.group(1))
     end = int(m.group(2) or start)
@@ -36,7 +37,7 @@ def main() -> int:
         "--hours-range",
         default="0-23",
         help="Hours of day interval expresses as 0-23 or 2",
-        type=parseNumList,
+        type=parseHourList,
     )
     args = parser.parse_args()
 
--- a/src/runner.py
+++ b/src/runner.py
@@ -3,9 +3,24 @@
 # Simple Script to replace cron for Docker
 
 import argparse
+import re
 import sys
+from datetime import datetime
 from subprocess import CalledProcessError, run
 from time import sleep, time
+
+
+def parseNumList(string):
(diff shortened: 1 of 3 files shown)


src/runner.py
```diff
@@ -3,11 +3,26 @@
...
 
 import argparse
+import re
 import sys
+from datetime import datetime
 from subprocess import CalledProcessError, run
 from time import sleep, time
 
 
+def parseNumList(string):
```
cooperlees, line 13:
```suggestion
def parseHourList(time_str: str) -> List[int]:
```
- Nit: Maybe think about the function name just to clearly state what it's doing ...
- Lets also type annotate since bandersnatch is pretty heavily time annotated (I didn't check if suggestions are right)

src/runner.py
```diff
@@ -17,31 +32,40 @@ def main() -> int:
...
+                except CalledProcessError as cpe:
+                    return cpe.returncode
+
+                run_time = time() - start_time
+                if run_time < args.interval:
+                    sleep_time = args.interval - run_time
+                    print(f"Sleeping for {sleep_time}s", file=sys.stderr)
+                    sleep(sleep_time)
+            else:
+                sleep(60)
```
cooperlees, line 68:
```suggestion
                sleep(args.interval)
```
Why default this to 60 and not use the supplied interval?

cooperlees (changes requested):
Thanks for this with a couple tunes. Happy to accept. Can we please:
- type annotate
- Unittest parseNumList/parseHourList please just incase people want to tweak it more in the future
  - This could help avoid regression if someone else comes along and makes changes

src/runner.py
mbacicc: I used a fixed 60 seconds there to avoid some impossible configurations. Like run with 7200 seconds and just a single hour in the provided interval. With 60 at least I'm sure it will check the current time every minute and run bandersnatch in every condition.

I'm sure there would better ways to do that, that was my idea in doing that.

src/runner.py
mbacicc: Done, I also added tests on the `parseHourList` function.
I had to call the test file as `test_docker_runner.py` in the same folder as the `runner.py` script for two reasons:

- `runner.py` is not a module, and thus the test file need to stay in the same folder to be imported correctly.
- if I called the test script as `test_runner.py` tox would be unhappy for the name confusion with the `test_runner.py` script in the repo root folder.

I hope I didn't create a mess with all these edits, wanted to contribute, but still learning how to work properly with python.

src/runner.py
mbacicc: I wanted to add tests for this, but I'm not an expert in mocking datetime and sleeps together. Would be nice to test that, too...

cooperlees: I can add these. Thanks for making some tests. Much appreciated.

cooperlees (approved):

Issue #1183, opened 2022-08-16
Add prettier markdown formatting to CI


cooperlees:
Lets add prettier markdown formatting to either the:
- doc build tox run
- pre-commit ci

Both work for me. Probably pre-commit since we use it already and most of our linting + formatting is there.

PR #1267, opened 2022-11-04
Add mdformat to pre-commit

RobertRosca:
https://github.com/pypa/bandersnatch/issues/1183 suggests adding "prettier" markdown formatter to pre-commit/ci, not sure if that means just any formatter to make markdown prettier or specifically the formatter called [prettier](https://prettier.io/) :stuck_out_tongue: 

This PR:

- Adds the [ExecutableBooks Markdown formatter](https://github.com/executablebooks/mdformat) to pre-commit. I chose this instead of prettier as for the reasons mentioned in the ['Why not use prettier instead?'](https://github.com/executablebooks/mdformat#why-not-use-prettier-instead) section of their readme
- Adds in a separate pre-commit workflow to execute pre-commit. Personally I prefer having these linting checks in a different workflow to the CI one, as this way the CI will still execute even if there are formatting issues, but I don't have a strong preference
- Makes required changes to existing markdown files (found by running `pre-commit run --all`) 

Closes https://github.com/pypa/bandersnatch/issues/1183

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,3 +1,4 @@
+---
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v4.3.0
@@ -37,6 +38,14 @@ repos:
     hooks:
       - id: flake8
         additional_dependencies: [flake8-bugbear]
+
+  - repo: https://github.com/executablebooks/mdformat
+    rev: 0.7.16
+    hooks:
+      - id: mdformat
+        additional_dependencies:
+          - mdformat-gfm
+          - mdformat-black
 
 ci:
   autofix_prs: false
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -71,7 +71,7 @@
 - Add bandersnatch command line help to the documentation main page `PR #920` - Thanks **ichard26**
 - Generate data-yanked tag in simple page `PR #931` - Thanks **happyaron**
 - Protect repository metadata from being trashed when disk is full `PR #962` - Thanks **happyaron**
-- Fix tox to used pinned requirements*.txt files for deps - `PR #1011` - Thanks **cooperlees**
+- Fix tox to used pinned requirements\*.txt files for deps - `PR #1011` - Thanks **cooperlees**
 
 ## Documentation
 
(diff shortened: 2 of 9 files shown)


.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,14 @@
...
+    branches: [main]
+  pull_request:
+
+jobs:
+  pre-commit:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: actions/setup-python@v3
+    - uses: pre-commit/action@v3.0.0
```
cooperlees, line 14: We already run pre-commit via it's CI offerings. I don't think we need this ?

cooperlees (changes requested):
Pure python and the other noted reasons are great reasons to use this. Thanks for sharing that this exists!

Lets just workout if we can remove the action or remove pre-commit CI - I think we can ...

cooperlees:
It seems mdformat also has [MyST support](https://github.com/executablebooks/mdformat#installing). Should we look into this since we use that render our docs?

.github/workflows/pre-commit.yml
RobertRosca: Aha yes, you're right, I forgot that pre-commit had their own CI integration.

I've dropped the commit that added this file.

RobertRosca:
> It seems mdformat also has [MyST support](https://github.com/executablebooks/mdformat#installing). Should we look into this since we use that render our docs?

I added the myst formatter, ran it, and it made a few minor changes to escape curly braces. One of them was in the docs, changing a fence `::{warning}` to `::\{warning}`, which seemed a bit fishy. I built the docs and that showed some warnings:

```
writing output... [100%] storage_options                                                                                                                                                 
/home/roscar/work/github.com/RobertRosca/bandersnatch/docs/mirror_configuration.md:114: WARNING: Pygments lexer name '\\{warning}' is not known
/home/roscar/work/github.com/RobertRosca/bandersnatch/docs/mirror_configuration.md:165: WARNING: Pygments lexer name '\\{note}' is not known
generating indices... genindex py-modindex done
```

There are some issues (https://github.com/executablebooks/mdformat-myst/issues/13) related to this, and it looks like the escaping is expected behaviour. So I'd recommend not adding this formatter in as it will would always attempt to reformat this fence syntax, causing issues with the docs.

cooperlees (approved):
Awesome then. Thanks for this and digging into everything.

PR #1281, opened 2022-11-16
add docker-compose example with webserver

tibeer:
This is an outcome of issue #1271.
Might help others.

Signed-off-by: Tim Beermann <beermann@osism.tech>


src/bandersnatch_docker_compose/docker-compose.yml
```diff
@@ -0,0 +1,24 @@
...
+    ports:
+      - "40080:80"
+    networks:
+      - bandersnatch_net
+
+networks:
+  bandersnatch_net:
+    driver: bridge
+    driver_opts:
+      com.docker.network.driver.mtu: 800
```
cooperlees, line 23: 800 if very low. How did you calculate that was needed?

I've never had to lower MTU on my docker networks, but I don't really do super heavy networking on any of my docker containers ...

cooperlees (changes requested):
Thanks for this. I'm sure it will be helpful.

I'm not a fan of the copy paste of files already in the repository. Could we maybe symlink them? So if someone updates one they "both" get updated?

Also, can you run `pre-commit` on the PR so it formats the markdown file.

src/bandersnatch_docker_compose/docker-compose.yml
tibeer: Good catch! I took the very low MTU to make sure that it definitely works. 1400 should also do the trick, but I would rather extend the README with hints regarding the numbers.

cooperlees (approved):
Many thanks for this and the explanations :) Hope this helps others who wish to do this. I use ansible to drive all my docker containers, so never really got into compose.

- I also don't use docker @ $dayJob ... so bit of a newb.

PR #1328, opened 2023-01-01
packages config for prerelease filter

TechCiel:
Make prerelease filter able to filter only specified packages.

May be useful for projects like `duckdb`, which publishes new prerelease for each commit.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@
 
 - Format with black `--preview` style `PR #1313`
   - I contribute to black and want to help find bugs ...
+- Add a `packages` config under `[filter_prerelease]` for prerelease filter to filter only specified packages. `PR #1328`
 
 ## Dropped Support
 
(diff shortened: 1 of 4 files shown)


cooperlees (changes requested):
Thanks. All looks good to me. Would you have time to add a unittest and show it working in https://github.com/pypa/bandersnatch/blob/main/src/bandersnatch/tests/plugins/test_prerelease_name.py please :D

If not, say so. I won't block, but it would be awesome if you could keep coverage up!

Please also add a Change Log entry to CHANGES.md

TechCiel:
@cooperlees , thanks for this quick review!

Changed the tests structure slightly to implement a proper unit test for this.

Commit amended.

cooperlees (approved):
Many thanks for the tests! Appreciate it - Will help ensure it stays working. Thanks.

PR #1335, opened 2023-01-03
Remove PURGE call on StalePage

### LINKED ISSUE #1334: Stop issuing PURGE requests on stale serial as PyPI requires authentication now
In https://github.com/pypa/bandersnatch/pull/57, we added code to automatically issue PURGE requests to stale endpoints. This was to work around a persistent issue w/ PyPI where caches become routinely stale and unusable. This hack worked for a good while, but now PyPI requires authentication for PURGE requests, flooding the logs with useless 401 errors.

```
❯ curl -X PURGE -i https://pypi.org/pypi/alchemize/json
HTTP/2 401 
server: Varnish
retry-after: 0
content-type: application/json
fastly-purge-auth-failure: 1
accept-ranges: bytes
date: Tue, 03 Jan 2023 00:03:15 GMT
x-varnish: 2329315070
via: 1.1 varnish
x-served-by: cache-yyz4563-YYZ
content-length: 53

{"msg":"Provided credentials are missing or invalid"}⏎   
```

See also: https://github.com/pypa/bandersnatch/pull/57#issuecomment-1367974037

cooperlees:
- This call now 401s so there is no benefit of calling it
- Also spams logs with noise - So lets remove before more people open issues

Test:
- See tests pass
  - Remove assertion of request being sent on StalePage as we no longer do ... That is this PRs purpose

Fixes #1334

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,7 +9,7 @@
 ## Dropped Support
 
 - Dropped LegacyVersion support for latest_name plugin `PR #1315`
-- No longer issue PURGE on StalePage `PR #1335`
+- No longer issue PURGE requests on StalePage exceptions as PyPI now requires authentication `PR #1335`
 
 # 6.0.1
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,7 @@
 ## Dropped Support
 
 - Dropped LegacyVersion support for latest_name plugin `PR #1315`
+- No longer issue PURGE on StalePage `PR #1335`
 
 # 6.0.1
 
--- a/src/bandersnatch/master.py
+++ b/src/bandersnatch/master.py
@@ -114,30 +114,10 @@ class Master:
             # really important check to achieve consistency and you should only
             # leave it out if you know what you're doing.
             if not got_serial or got_serial < required_serial:
-                logger.debug(
-                    f"Expected PyPI serial {required_serial} for request {path} "
-                    + f"but got {got_serial}"
-                )
-
-                # HACK: The following attempts to purge the cache of the page we
-                # just tried to fetch. This works around PyPI's caches sometimes
-                # returning a stale serial for a package. Ideally, this should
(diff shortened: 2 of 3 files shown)


src/bandersnatch/master.py
```diff
@@ -114,30 +114,10 @@ async def check_for_stale_cache(
...
-                    logger.warning(
-                        "Got an error when attempting to clear the cache", exc_info=True
-                    )
-
                 raise StalePage(
                     f"Expected PyPI serial {required_serial} for request {path} "
-                    + f"but got {got_serial}. "
-                    + "HTTP PURGE has been issued to the request url"
+                    + f"but got {got_serial}. We can no longer issue a PURGE. "
+                    + "Report issue to PyPA Warehouse GitHub if it persists ..."
```
ichard26, line 120:
```suggestion
                    + f"but got {got_serial}. We can no longer issue a PURGE. "
                    + "Report issue to pypa/warehouse on GitHub if it persists ..."
```
Ideally we'd link to a pre-existing issue, otherwise people will almost never report persistent stale responses on Warehouse. TBF the likelihood is already tiny, but a direct link to an issue would probably help. I don't feel strongly about this though.

CHANGES.md
```diff
@@ -9,6 +9,7 @@
 ## Dropped Support
 
 - Dropped LegacyVersion support for latest_name plugin `PR #1315`
+- No longer issue PURGE on StalePage `PR #1335`
```
ichard26, line 12:
```suggestion
- No longer issue PURGE requests on StalePage exceptions as PyPI now requires authentication `PR #1335`
```

ichard26 (approved):

src/bandersnatch/master.py
cooperlees: I don't want a link that could go stale and it at the end of the day only hurts them by not reporting ...

ichard26: That's fair!

PR #1340, opened 2023-01-06
Storage.scandir() method for replacing glob()s

TechCiel:
Implemented a new `Storage.scandir(self, path)` API, used to replace `glob("**")`s for large directory listing.

### DIFF
--- a/src/bandersnatch/simple.py
+++ b/src/bandersnatch/simple.py
@@ -6,8 +6,6 @@ from pathlib import Path
 from typing import TYPE_CHECKING, Any, Dict, List, NamedTuple, Optional, Union
 from urllib.parse import urlparse
 
-from packaging.utils import canonicalize_name
-
 from .package import Package
 
 if TYPE_CHECKING:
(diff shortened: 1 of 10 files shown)


src/bandersnatch/simple.py
```diff
@@ -114,7 +111,11 @@ def get_simple_dirs(self, simple_dir: Path) -> List[Path]:
             # We are using index page directory hashing, so the directory
             # format is /simple/f/foo/.  We want to return a list of dirs
             # like "simple/f".
-            subdirs = [simple_dir / x for x in simple_dir.iterdir() if x.is_dir()]
+            subdirs = sorted(
+                simple_dir / str(x.name)
```
cooperlees, line 115: Isn't Path().name already a str? Maybe not for some of the storage plugins or protocol you just added? Why you casting to a str everywhere?

src/bandersnatch_storage_plugins/s3.py
```diff
@@ -331,6 +331,15 @@ def mkdir(
             path = self.PATH_BACKEND(path)
         path.joinpath(self.PATH_BACKEND.keep_file).touch()
 
+    def scandir(self, path: PATH_TYPES) -> Generator[S3DirEntry, None, None]:
+        """Read entries from the provided directory"""
+        if not isinstance(path, self.PATH_BACKEND):
+            path = self.PATH_BACKEND(path)
+        for p in path._accessor.scandir(path):
+            if p.name == ".s3keep":
+                continue
```
cooperlees, line 340: Is this special to s3 - Can we link to why we skip this for less s3 versed people? Or just a comment.

cooperlees (approved):
LGTM - Many thanks. Went through - I've never done a protocol class. TIL. Unit testing looks good and CI passes - But I don't think we run a second run very well to test upgrades perfectly but it's pretty high coverage so I am confident here.

Just some small comments - If you don't feel they are worth it just say so and I'll look to merge. This will be a great perf improvement.

src/bandersnatch/simple.py
TechCiel: Because `os.DirEntry.name` have the type `AnyStr = TypeVar('AnyStr', bytes, str)`, we have to `str()` here.

src/bandersnatch_storage_plugins/s3.py
TechCiel: Yes, but for swift there's `.swiftkeep`. Actually, I should have used `self.PATH_BACKEND.keep_file` here to align with the `mkdir()`, which have explained just lines above.

TechCiel: Just fixed this. ~~Oh no the actions got started again...~~

PR #1341, opened 2023-01-07
Move Docker to 3.11

cooperlees:
We work in 3.11 so might as well start building docker in the faster available Python Runtime.

Please report any bugs!

Will look to go >= 3.10 later this year ... And move code to take advantage of that if we can.


cooperlees:
Unfortunately CI won't test anything here until we merge to main - So I might just merge and forward fix here once unit tests etc. pass ...

PR #1347, opened 2023-01-12
Add parameter to allow passing "--force-check" to bandersnatch.

tibeer:
When running this as a mirror container, it might be desired to always update.

Signed-off-by: Tim Beermann <beermann@osism.tech>


cooperlees (approved):
Sure thing. This is helpful.

Long term: I wonder if we could import from bandersnatch all the argparse option for mirror and just get them for free and use that generate the CLI args ...

cooperlees:
FWIW - the `latest` docker container will get this on merge. I do need to do a release tho.

Issue #1386, opened 2023-02-25
Exception with UTF-16 encoded requirements.txt file


francescocaponio:
I have this exception 
```
2023-02-25 22:29:51,784 INFO: considering /requirements/*********-requirements.txt (allowlist_name.py:114)
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/usr/local/lib/python3.11/site-packages/bandersnatch/main.py", line 231, in <module>
    exit(main())
         ^^^^^^
  File "/usr/local/lib/python3.11/site-packages/bandersnatch/main.py", line 227, in main
    return asyncio.run(async_main(args, config))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/asyncio/runners.py", line 190, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/asyncio/base_events.py", line 653, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/bandersnatch/main.py", line 192, in async_main
    return await bandersnatch.mirror.mirror(config)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/bandersnatch/mirror.py", line 963, in mirror
    mirror = BandersnatchMirror(
             ^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/bandersnatch/mirror.py", line 203, in __init__
    super().__init__(master=master, workers=workers)
  File "/usr/local/lib/python3.11/site-packages/bandersnatch/mirror.py", line 42, in __init__
    self.filters = LoadedFilters(load_all=True)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/bandersnatch/filter.py", line 151, in __init__
    self._load_filters(self.ENTRYPOINT_GROUPS)
[...]

PR #1387, opened 2023-02-26
Fixed import of packages from requirement files with UTF-16 encoding

francescocaponio:
Solves #1386

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -4,7 +4,7 @@ max_line_length = 88
 # E722 is a duplicate of B001.
 # P207 is a duplicate of B003.
 # W503 is against PEP8
-ignore = E722, P207, W503
+ignore = E722, P207, W503, E203
 max-complexity = 20
 exclude =
     build,
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,7 +9,7 @@
 
 ## Bug Fixes
 
-- Fixed import of packages from requirement files with UTF-16 encoding `PR #1386`
+- Fixed import of packages from requirement files with UTF-16 encoding `PR #1387`
 
 # 6.1.0
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,10 @@
 - Move Docker to build in 3.11 `PR #1341`
 - Add "--force-check" parameter to runner.py `PR #1347`
 
+## Bug Fixes
+
+- Fixed import of packages from requirement files with UTF-16 encoding `PR #1386`
+
 # 6.1.0
 
 ## New Features
--- a/src/bandersnatch/tests/plugins/test_allowlist_name.py
+++ b/src/bandersnatch/tests/plugins/test_allowlist_name.py
@@ -572,3 +572,37 @@ requirements =
         # Check that the package in the last file, excluded
         # from the glob is not considered
         self.assertNotIn("baz", mirror.packages_to_sync)
+
+    def test__filter__requirements__utf16__encoding(self) -> None:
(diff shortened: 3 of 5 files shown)


ichard26 (commented):
This doesn't seem to do anything? It's just opening the file twice without saving the read contents anywhere.

It's probably better to just loudly error in the logs if the filesystem/OS default encoding doesn't work. Maybe we could be smarter about which one we use, but that's beyond my expertise.

francescocaponio:
Actually I searched a lot and there is no way to know before opening a file which is the encoding.

The try except block tries to open it first with UTF-8 encoding (the more common), if it fails with UnicodeDecode Exception, it opens it with UTF-16, maybe the UTF-16 open in the except block is unuseful, since it will be later opened again to be read. That could be avoided or used in another try except block, if it would be the case in the future (I would hope not). 

Windows powershell exports the requirements file in UTF-16, while bash does it in UTF-8.

As I was explaining ih the #1386, bandersnatch process is quite unattended, I don't want it to be stuck if a wrong encoded file goes in the requirements folder.

cooperlees (changes requested):
Thanks for attempting a PR. I commented on the issue and I feel we should just use [charset-normalizer](https://pypi.org/project/charset-normalizer/) and detect the encoding and default to `utf-8` if it fails to detect so we support all encoding that `pip` does with requirements files.

francescocaponio:
I didn't think about PEP rules and actually never noticed I had some UTF-16 encoded requirements until I provided them to bandersnatch. 
Pip installs them without any problem:
image
pip uses the function auto_decode to open files: https://github.com/pypa/pip/blob/61e85bf27996a9a73170e353374f13b742ad2058/src/pip/_internal/utils/encoding.py#L20
maybe we can use the same too?

cooperlees:
This is fine that windows does this and pip supports it. Like add pip like support here to bandersnatch please and that will save us ever having to come back to it. I'm all for adding it, but i think it's worth the little bit of extra effort. Thank for diffing in and working it all out.

francescocaponio:
Ok, sorry to bother you.
I have the edited version, but I'm unable to pass the pre-commit scripts. Can you help me?
The problem is happening here:
image
black adds a whitespace, and flake8 is unhappy about it:
image

Did I do something wrong?

BTW pytest is passing.

.flake8
```diff
@@ -4,7 +4,7 @@ max_line_length = 88
 # E722 is a duplicate of B001.
 # P207 is a duplicate of B003.
 # W503 is against PEP8
-ignore = E722, P207, W503
+ignore = E722, P207, W503, E203
```
cooperlees, line 7: Thanks - Wonder how we never hit this ... Just must have never done it.

CHANGES.md
```diff
@@ -7,6 +7,10 @@
 - Move Docker to build in 3.11 `PR #1341`
 - Add "--force-check" parameter to runner.py `PR #1347`
 
+## Bug Fixes
+
+- Fixed import of packages from requirement files with UTF-16 encoding `PR #1387`
```
cooperlees, line 12:
```suggestion
- Fixed reading requirement files with UTF-16 encoding supporting all encoding pip does `PR #1387`
```
We kind of do more than just UTF-16 now ... But I can fix later if you don't get to it.

cooperlees (approved):
LGTM - Thanks. This will indeed support all pip supports.

I might look into or open a follow up issues to see if we can remove this copy pasted code and move to a library that will be kept up to date for us. But this is fine for now and does more than only adding UTF-16

PR #1442, opened 2023-04-27
Use the configurable digest_name for simple JSON

### LINKED ISSUE #1440: package json digest dict mapped to simple json hashes dict causes pip >23 to fail
blake2b_256 has been added to the digest dict in package json, pip does not know how to handle this hash name. Causing and exception when trying to install packages with pip23 or later. 

If you check https://pypi.org/simple/pip against https://pypi.org/simple/pip/23.0.1/json
the digest dict does not contain the same data as the hashes dict. 

```
$ pip install --force pip
Looking in indexes: https://<pypi mirror>/simple/
Collecting pip
  Using cached https://<pypi mirror>/packages/07/51/2c0959c5adf988c44d9e1e0d940f5b074516ecc87e96b1af25f59de9ba38/pip-23.0.1-py3-none-any.whl (2.1 MB)
ERROR: Unknown hash name: blake2b_256
```

This is caused by how bandersnatch discovers the hashes that should be applied to /simple/
https://github.com/pypa/bandersnatch/blob/f99542f17adcc5c35ddfbb62b03a84d43122793a/src/bandersnatch/simple.py#L192
```
def generate_json_simple_page(
        ...
        # Add release files into the JSON dict
        for r in release_files:
            package_json["files"].append(
                {
                    "filename": r["filename"],
                    "hashes": {
                        digest_name: digest_hash
                        for digest_name, digest_hash in r["digests"].items()
                    },
                    "requires-python": r.get("requires_python", ""),
                    "url": self._file_url_to_local_url(r["url"]),
                    "yanked": r.get("yanked", False),
                }
            )
```

haatveit:
PEP 691 specifies that each file's hashes dictionary must contain only
hash algorithms that can be passed, without arguments, to hashlib.new().
PyPI has started using a "blake2b_256" key for digests, which is blake2b
with the digest size set to 32 bytes, and makes it so that contents of
"digests" can't just be reused for the hashes dictionary.

This commit uses the value of configuration option digest_name to pick
a single supported key/value pair from the digests for use in the hashes
dictionary.

### DIFF
--- a/src/bandersnatch/simple.py
+++ b/src/bandersnatch/simple.py
@@ -36,7 +36,7 @@ def get_format_value(format: str) -> SimpleFormat:
     try:
         return SimpleFormat[format.upper()]
     except KeyError:
-        valid_formats = [v.name for v in SimpleFormat].sort()
+        valid_formats = sorted([v.name for v in SimpleFormat])
         raise InvalidSimpleFormat(
             f"{format.upper()} is not a valid Simple API format. "
             + f"Valid Options: {valid_formats}"
--- a/src/bandersnatch/simple.py
+++ b/src/bandersnatch/simple.py
@@ -42,7 +42,7 @@ class InvalidSimpleFormat(KeyError):
     pass
 
 
-class InvalidDigestFormat(KeyError):
+class InvalidDigestFormat(ValueError):
     """We don't have a valid digest choice from configuration"""
 
     pass
@@ -65,7 +65,7 @@ def get_digest_value(digest: str) -> SimpleDigest:
     except KeyError:
         valid_digests = sorted([v.name for v in SimpleDigest])
         raise InvalidDigestFormat(
-            f"{digest.upper()} is not a valid Simple API file hash digest. "
+            f"{digest} is not a valid Simple API file hash digest. "
             + f"Valid Options: {valid_digests}"
         )
 
--- a/src/bandersnatch/simple.py
+++ b/src/bandersnatch/simple.py
@@ -23,11 +23,27 @@ class SimpleFormat(Enum):
     JSON = auto()
 
 
+class SimpleDigests(NamedTuple):
+    sha256: str
+    md5: str
+
+
+class SimpleDigest(str, Enum):
+    SHA256 = "sha256"
+    MD5 = "md5"
+
+
 logger = logging.getLogger(__name__)
 
 
 class InvalidSimpleFormat(KeyError):
     """We don't have a valid format choice from configuration"""
+
+    pass
+
+
+class InvalidDigestFormat(KeyError):
+    """We don't have a valid digest choice from configuration"""
 
     pass
 
@@ -40,6 +56,17 @@ def get_format_value(format: str) -> SimpleFormat:
         raise InvalidSimpleFormat(
             f"{format.upper()} is not a valid Simple API format. "
             + f"Valid Options: {valid_formats}"
+        )
+
+
+def get_digest_value(digest: str) -> SimpleDigest:
+    try:
+        return SimpleDigest[digest.upper()]
+    except KeyError:
+        valid_digests = sorted([v.name for v in SimpleDigest])
+        raise InvalidDigestFormat(
+            f"{digest.upper()} is not a valid Simple API file hash digest. "
+            + f"Valid Options: {valid_digests}"
         )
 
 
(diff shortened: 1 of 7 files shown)


mbeno:
Maybe using same logic here as for the html simple page would be good enough?
```
"hashes": {
    self.digest_name = r.["digests"][digest_name]         
},
```

src/bandersnatch/simple.py
```diff
@@ -1,3 +1,4 @@
+import hashlib
```
cooperlees, line 1: This isn't used right? Lets remove.

cooperlees (changes requested):
Thanks for this. It makes sense, expecially if PyPI is doing this. Lets use the config chosen hash algorithm only for now to be as "simple" as possible.

I'd also expect some tests to need fixtures updates here or I suck at writing tests. I've enabled them to run, lets fix what fails to please.

haatveit:
>I'd also expect some tests to need fixtures updates here or I suck at writing tests. I've enabled them to run, lets fix what fails to please.

Yeah this would indeed fail tests - I totally forgot to rerun them after my initial change, sorry.

----

I've basically mimicked your approach with the formats for digest selection in cf8e7cd7c714c61facfbe377559b58c83a83ac4c, and then updated tests to work with these changes. Just let me know if anything there is still iffy.


So far the digest_name has been validated in the configuration module:

https://github.com/pypa/bandersnatch/blob/f95e0500b24a844743f008cb64162e4de6257153/src/bandersnatch/configuration.py#L133-L142

Pretty redundant at the moment so I guess I should address this too.

src/bandersnatch/simple.py
```diff
@@ -23,6 +29,16 @@ class SimpleFormat(Enum):
     JSON = auto()
 
 
+class SimpleDigests(NamedTuple):
+    sha256: str
+    md5: str
```
cooperlees, line 34: Not for this PR - but I wonder how we can support any hashname moving forward and just have the code adapt if PyPI change / offers new hashes and people want to move to it. I might open a Wishlist Issue.

I might in the short term add an exception / error message explaining we need a code diff for a new hash ...

src/bandersnatch/simple.py
```diff
@@ -32,17 +48,34 @@ class InvalidSimpleFormat(KeyError):
...
+
+    pass
+
+
 def get_format_value(format: str) -> SimpleFormat:
     try:
         return SimpleFormat[format.upper()]
     except KeyError:
-        valid_formats = [v.name for v in SimpleFormat].sort()
+        valid_formats = sorted([v.name for v in SimpleFormat])
```
cooperlees, line 61: Just out of curiosity, why did this change need to happen? Did you need a new list for some reason, I believe this is slower (not that it really matters here)

cooperlees (approved):
LGTM - Thanks for the persistence in getting all the CI happy. Appreciate it. Will merge cut a release today

I need to check the code and see if a `bandersnatch mirror --force-check` will be enough to cause every PEP691 JSON to be regenerated ... If not I'll do a PR adding an option to `--force-simple-rebuild`

cooperlees:
Reading the code, it does indeed look like `--force-check` is enough to regenerate all the JSON pages. It also seems a `bandersnatch sync` would fix specific packages too ... I will put in release notes.

src/bandersnatch/simple.py
haatveit: I think this should be doable if one creates the named tuple and enum with a functional approach rather than the static classes that were defined in this PR. Some decisions would definitely need to be made on how to check what algorithms are available.

src/bandersnatch/simple.py
haatveit: I didn't consider speed to be honest, but that is good to know. `valid_formats` would never get the intended value here because `list.sort()` returns `None` after modifying the object in-place.

haatveit:
Thank you very much for reviewing and merging this!

src/bandersnatch/simple.py
cooperlees: Yup, you are right here. I missed it. Thanks for the fix.

src/bandersnatch/simple.py
cooperlees: Totally, I feel we just pull what ever the config allows/asks for and error if it's not in the JSON metadata we pull. I might open a cleaerer issue / do it one day if I get time. I like to avoid static code that needs a code push to give people potential future functionality (e.g. a new hash to be supported) where ever possible.

mbeno:
Thanks @cooperlees and @haatveit for helping resolve this issue so quickly! ❤️

cooperlees:
Awesome. Thanks for letting us know it fixes your mirror.

PR #1450, opened 2023-05-02
fix: draft fix for sync vulnerability

caesarw:
This pull request fixes a bug where bandersnatch did not properly handle deletion of a project/all releases of a project. 

The vulnerability occurs when bandersnatch failed to generate the correct Simple Index and notify the client (which is usually pip or poetry) that some packages have already been deleted by the upstream repo (PyPI.org). 

For example, **_project A_** was completely removed by PyPI (which means all associated releases or the project on PyPI is removed). However, third-party PyPI mirrors using unpatched version of bandersnatch not only still keeps the source and the wheel of **_project A_**, and would still make **_project A_** downloadable and installable by client who decided to use that mirror.


cooperlees:
This needs a lot of work, please reopen a new PR if you plan to finish this. Thanks for attempting tho!

PR #1547, opened 2023-09-13
fix verify behind proxy and allow skip sha256 check

LeoQuote:
1. fix bug: proxy is not set when calling `verify`, setting it to let proxy work
2. skip sha256 check in storage other than file system, could be enforced with `--force-sha256-check` argument. This is quite expensive, both in time and finance, download all packages from s3 could cost much.

### DIFF
--- a/src/bandersnatch/main.py
+++ b/src/bandersnatch/main.py
@@ -96,6 +96,12 @@ def _verify_parser(subparsers: argparse._SubParsersAction) -> None:
         default=0,
         help="# of parallel iops [Defaults to bandersnatch.conf]",
     )
+    v.add_argument(
+        "--force-sha256-check",
+        action="store_true",
+        default=False,
+        help="force sha256 check on all storage backend",
+    )
     v.set_defaults(op="verify")
 
 
@@ -155,6 +161,7 @@ async def async_main(args: argparse.Namespace, config: ConfigParser) -> int:
             config.get("mirror", "master"),
             config.getfloat("mirror", "timeout"),
             config.getfloat("mirror", "global-timeout", fallback=None),
+            config.get("mirror", "proxy"),
         ) as master:
             return await bandersnatch.delete.delete_packages(config, args, master)
     elif args.op.lower() == "verify":
--- a/src/bandersnatch/master.py
+++ b/src/bandersnatch/master.py
@@ -52,6 +52,15 @@ class Master:
             logger.error(err)
             raise ValueError(err)
 
+    def _build_proxy_connector(self) -> ProxyConnector | None:
+        """Build a ProxyConnector if a proxy is set in the config"""
+        if not self.proxy:
+            connector = self._check_for_socks_proxy()
+            return connector
+
+        logger.debug(f"Creating a ProxyConnector to use {self.proxy}")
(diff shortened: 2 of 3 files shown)


src/bandersnatch/verify.py
```diff
@@ -117,6 +117,7 @@ async def verify(
     args: argparse.Namespace,
     executor: concurrent.futures.ThreadPoolExecutor | None = None,
     releases_key: str = "releases",
+    sha256_check: bool = True
```
cooperlees, line 120: This should be more generic as I'm sure we'll move from sha256 one day ... But this is outside this PR

src/bandersnatch/master.py
```diff
@@ -139,7 +148,6 @@ async def get(
             await self.check_for_stale_cache(path, required_serial, got_serial)
             yield r
 
-    # TODO: Add storage backend support / refactor - #554
```
cooperlees, line 142: Thanks. Great to know it works on s3.

src/bandersnatch/master.py
```diff
@@ -52,6 +52,15 @@ def __init__(
             logger.error(err)
             raise ValueError(err)
 
+    def _build_proxy_connector(self) -> ProxyConnector | None:
+        """Build a ProxyConnector if a proxy is set in the config"""
+        if not self.proxy:
```
cooperlees, line 57: Do we verify self.proxy is valid proxy URI aywhere or just leave that to when we go to use it via socks_proxy or aiohttp http proxy support?
- Just food for thought here - Wondering if we should do better ...

src/bandersnatch/main.py
```diff
@@ -96,6 +96,12 @@ def _verify_parser(subparsers: argparse._SubParsersAction) -> None:
         default=0,
         help="# of parallel iops [Defaults to bandersnatch.conf]",
     )
+    v.add_argument(
+        "--force-sha256-check",
+        action="store_true",
+        default=False,
+        help="force sha256 check on all storage backend",
```
cooperlees, line 103: Can we say it defaults to on for filesystem backend
```suggestion
        help="Force sha256 check on all storage backends. POSIX filesystems default to checking.",
```

cooperlees (changes requested):
Looks good. Let's fix CI ... would love to try and cover the new code with tests somehow

`_build_proxy_connector` should be easily testable.

LeoQuote:
@cooperlees Thanks for your review, this pr is still under construction, I'll mark this ready when all the CI goes green.

src/bandersnatch/verify.py
LeoQuote: yes, but I didnt find a better name, will try later 😄

cooperlees:
Sure thing, don't worry to much about the sha256 CLI arg name. We can change in the future ...

cooperlees:
Any plans to finish this?

LeoQuote:
@cooperlees sorry but I cannot finish this pr, I've pushed every commit of my work, I have given up on trying using bandersnatch with s3, or in our team at least, we're currently using a PyPI mirror provided by our vendor, which is working pretty well, thanks for your work and review!

Issue #1433, opened 2023-04-18
Improve mirror filtering for latest releases


kefir-:
As described in the docs, bandersnatch can filter mirror to only include the latest releases:

https://bandersnatch.readthedocs.io/en/latest/filtering_configuration.html#keep-only-latest-releases

But this can introduces some problems, and I think it boils down to a definition of what exactly the latest releases are. Currently, bandersnatch appears to sort all releases by version number:

https://github.com/pypa/bandersnatch/blob/main/src/bandersnatch_filter_plugins/latest_name.py#L52

and then it selects the versions to keep. But this is a problem, because there may be two major releases under maintenance, and if you don't fetch everything you might miss security fixes for the old major release. For example, if the chronological release order looks like this:

1.0.0
1.1.0
1.2.0
1.3.0
1.3.1
1.3.2
2.0.0
2.0.1
1.3.3
2.0.2
2.0.3
2.0.4
2.0.5
2.0.6
2.0.7
1.3.4 critical patch
2.0.8
2.0.9

If keep is set to 10 and we run bandersnatch, then v1.3.4 will not be fetched, even if it was released yesterday. So instead of sorting by version numbers, it might make sense to instead keep the latest 10 versions listed under Release history, eg https://pypi.org/project/bandersnatch/#history or https://pypi.org/rss/project/bandersnatch/releases.xml. 

Also, a more sensible setting might be to add a different option to fetch all updates from the last X days/months (falling back to the latest release if it was even older). This seems less arbitrary than a release version count of for example 20, which could be 20 releases that are all 10 years old for one project, and 20 releases within only the last week for another project.

cooperlees:
A PR is welcome to add a configurable "chronological" latest releases addition to the plugin. I'll only accept it configurable and with unittests added showing how this works with documentation showing how to use and stating how it differs.

bandersnatch uses the "JSON" and "simple" APIs from PyPI to get metadata today. I guess you'll have to use the `upload_time_iso_8601` of each release to find the last X chronological versions. I don't think we need to pull in the XML as well to do this. I'd like to avoid adding more HTTPS calls.

Thanks!

PR #1594, opened 2023-11-04
#1433 improve release filtering by adding time based option

miicha:
Time based sorting can be used by setting `sort_by = time` together with `keep`

```
[latest_release]
keep = 3
sort_by = time
```
implements https://github.com/pypa/bandersnatch/issues/1433

### DIFF
--- a/docs/filtering_configuration.md
+++ b/docs/filtering_configuration.md
@@ -237,9 +237,12 @@ enabled =
 
 [latest_release]
 keep = 3
+sort_by = [version/time]
 ```
 
 By default, the plugin does not filter out any release. You have to add the `keep` setting.
+The default is to sort by `version` number (parsed according to semantic versioning).
+As an alternative, `time` can be used to sort releases chronologically by upload time and select the last `keep` ones.
 
 You should be aware that it can break requirements. Prereleases are also kept.
 
--- a/docs/filtering_configuration.md
+++ b/docs/filtering_configuration.md
@@ -241,7 +241,7 @@ enabled =
 
 [latest_release]
 keep = 3
-sort_by = [version/time]
+sort_by = [version|time]
 ```
 
 By default, the plugin does not filter out any release. You have to add the `keep` setting.
--- a/src/bandersnatch_filter_plugins/latest_name.py
+++ b/src/bandersnatch_filter_plugins/latest_name.py
@@ -38,6 +38,10 @@ class LatestReleaseFilter(FilterReleasePlugin):
             sort_by = self.configuration["latest_release"]["sort_by"]
             if sort_by in ["time", "version"]:
                 self.sort_by = sort_by
+            else:
+                logger.debug(
+                    "sort_by only allows 'time' and 'version', and not '{}'".format(sort_by)
+                )
             logger.info(
                 f"Initialized latest releases plugin with sort_by={self.sort_by}"
             )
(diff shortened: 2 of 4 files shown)


miicha:
pre-commit.ci autofix

src/bandersnatch_filter_plugins/latest_name.py
```diff
@@ -32,6 +34,17 @@ def initialize_plugin(self) -> None:
             return
         if self.keep > 0:
             logger.info(f"Initialized latest releases plugin with keep={self.keep}")
+        try:
+            sort_by = self.configuration["latest_release"]["sort_by"]
+            if sort_by in ["time", "version"]:
+                self.sort_by = sort_by
```
cooperlees, line 40: We should probably at least log.debug if someone sets a bad config option + state we're sticking with default of version sorting

src/bandersnatch_filter_plugins/latest_name.py
```diff
@@ -32,6 +34,17 @@ def initialize_plugin(self) -> None:
...
+            logger.info(
+                f"Initialized latest releases plugin with sort_by={self.sort_by}"
+            )
+        except KeyError:
+            return
+        except ValueError:
+            return
```
cooperlees, line 47: Lets log.error's here ... Unless something upstream does ...

What can throw a ValueError here?

cooperlees (changes requested):
Thanks. Looks good to me.

Lets just add some logging to make future debugging possibly easily + documentation please:
- Docs for filtering are here: https://github.com/pypa/bandersnatch/blob/main/docs/filtering_configuration.md#keep-only-latest-releases

src/bandersnatch_filter_plugins/latest_name.py
miicha: this is just copy paste from above (self.keep = int(self.configuration["latest_release"]["keep"])). I did not really think about possible errors, since `keep` is handled the same way...

src/bandersnatch_filter_plugins/latest_name.py
miicha: added a debug log warning

cooperlees:
pre-commit.ci autofix

src/bandersnatch_filter_plugins/latest_name.py
cooperlees: It's an int, so if a str that is not a number is set the ValueError can occur ... Blindly copy pasting without thinking about it should be done sparingly ...

miicha: Oups, missed that...
What about the KeyError? As far as I understand, nothing should be looged if some optional setting is not in the config file.
I would remove the ValueError and hope that the PR can then be merged:-)

docs/filtering_configuration.md
```diff
@@ -241,9 +241,12 @@ enabled =
 
 [latest_release]
 keep = 3
+sort_by = [version/time]
```
cooperlees, line 244:
```suggestion
sort_by = [version|time]
```

cooperlees (approved):
Welcome to the mix filter by upload time ...

FELIXrobust (approved):

FELIXrobust (approved):

FELIXrobust (approved):

Issue #1599, opened 2023-11-06
Latest x packages breaks pinned version functionality


miicha:
## My usecase
- I'd like to have a slim mirror for offline usage based on a selection of a few packages
- packages are initially given in a requirmentss.txt and externally expanded to a requirements.txt with all dependencies (including version range)
- To keep the required storage low, I'd like to fetch max X versions while honoring the pinned versions given in the requirements.txt

## Problem
Because all filters are applied in a loop in `package.py`, the last X versions plugin wins and older required versions are not downloaded:
```
if not all(plugin.filter(release_data) for plugin in release_filters):
    del self.releases[version]
```

## Possible solution
- The pinned filter should have priority and "pre-filter" the available versions
- the next step is not yet clear to me, but I can see 3 options:
    1. afterwards we can either filter the remaining versions to keep only X, but this is problematic because we don't know if different packages have different version ranges since we would need to keep at least one version from each range
    2. alternatively last X version filtering might be suppressed for packages that have pinned versions / version range leading to more packages to be downloaded
    3. the pinned versions filtering itself takes care of fetching only the newest version from each given range (would be my favorite solution)

Since I need this functionality quite urgently, I would create a "work in progress" PR with solution 2. If you have some additional ideas or requirements, I would adjust it accordingly.

cooperlees:
Thanks for proposing and looking into this. We've never prioritized getting plugin's ordering correct so we apply correctly as it really increases the complexity of the plugin system, as you're seeing.

I like the sounds of checking if there are specific per package version ranges etc. and exempting them from the keep option.

If you could prove this all works with unit tests it will increase my ability to merge this efficiently. I'd even accept adding this to the integration tests we run too.
- that is: https://github.com/pypa/bandersnatch/blob/main/test_runner.py when `TOXENV=INTEGRATION` is set ...

PR #1601, opened 2023-11-07
Latest and version pinning clash

miicha:
Make sure that pinned versions defined with `project_requirements_pinned` plugins are downloaded by disabling additional filter if version (range) is specified.

implements https://github.com/pypa/bandersnatch/issues/1599

### DIFF
--- a/src/bandersnatch/tests/plugins/test_allowlist_name.py
+++ b/src/bandersnatch/tests/plugins/test_allowlist_name.py
@@ -409,15 +409,12 @@ requirements =
     def test__filter__matches__release_latest(self) -> None:
         with open(Path(self.tempdir.name) / "requirements.txt", "w") as fh:
             fh.write("""\
-#    This is needed for workshop 1
-#
 foo==1.2.0             # via -r requirements.in
 """)
 
         mock_config(f"""\
 [mirror]
 storage-backend = filesystem
-workers = 2
 
 [plugins]
 enabled =
--- a/src/bandersnatch/tests/plugins/test_allowlist_name.py
+++ b/src/bandersnatch/tests/plugins/test_allowlist_name.py
@@ -406,6 +406,43 @@ requirements =
 
         self.assertEqual({"1.2.0": {}}, pkg.releases)
 
+    def test__filter__matches__release_latest(self) -> None:
+        with open(Path(self.tempdir.name) / "requirements.txt", "w") as fh:
+            fh.write("""\
+#    This is needed for workshop 1
+#
+foo==1.2.0             # via -r requirements.in
+""")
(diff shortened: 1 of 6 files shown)


miicha:
pre-commit.ci autofix

src/bandersnatch_filter_plugins/allowlist_name.py
```diff
@@ -202,6 +202,17 @@ def _determine_filtered_package_requirements(self) -> list[Requirement]:
...
 
+    def pinned_version_exists(self, metadata: dict) -> bool:
+        name = canonicalize_name(metadata["info"]["name"])
+        for requirement in self.allowlist_release_requirements:
+            if name != requirement.name:
+                continue
+            elif len(requirement.specifier) > 0:
+                return True
+            else:
+                return False
```
cooperlees, line 213:
```suggestion
            return len(requirement.specifier) > 0
```
Just a NIT for less lines

src/bandersnatch/tests/plugins/test_allowlist_name.py
```diff
@@ -406,6 +406,43 @@ def test__filter__matches__release(self) -> None:
 
         self.assertEqual({"1.2.0": {}}, pkg.releases)
 
+    def test__filter__matches__release_latest(self) -> None:
+        with open(Path(self.tempdir.name) / "requirements.txt", "w") as fh:
+            fh.write("""\
+#    This is needed for workshop 1
+#
```
cooperlees, line 413: Hope workshop 1 is fun :)

src/bandersnatch/tests/plugins/test_allowlist_name.py
```diff
@@ -406,6 +406,43 @@ def test__filter__matches__release(self) -> None:
...
+foo==1.2.0             # via -r requirements.in
+""")
+
+        mock_config(f"""\
+[mirror]
+storage-backend = filesystem
+workers = 2
```
cooperlees, line 420: Is there a reason two have 2 workers here? (It's fine, just interested)

cooperlees (approved):
LGTM - Thanks for making this. Makes sense to me.

src/bandersnatch/tests/plugins/test_allowlist_name.py
miicha: again just copy paste from the unit test above (removed with last commit)

src/bandersnatch/tests/plugins/test_allowlist_name.py
miicha: unfortunately not --> removed ;)

PR #1653, opened 2024-01-26
Improve + Fix Bandersnatch Docker Compose Setup: Volume Configuration + HTTPS Support

benjisho:
This pull request introduces significant enhancements to the Bandersnatch Docker Compose setup, focusing on improved security through HTTPS support and optimized volume configuration. The main points of this update are as follows:

1. **Bind Mounted Volume for Nginx Configuration**: 
   - Implemented a bind-mounted volume for the Nginx configuration file (`nginx.conf`) - allows users to easily modify the Nginx configuration without rebuilding the Docker image.
   - Updated the `docker-compose.yml` to correctly reference the `nginx.conf` file located in the `banderx` directory.

2. **Added HTTPS Support**:
   - Integrated optional HTTPS support within the Nginx configuration.
   - Added commented-out sections in both the `docker-compose.yml` and `nginx.conf` files for enabling HTTPS. Users can now simply uncomment these sections to activate HTTPS.
   - Written instructions in the `README.md` on how to enable HTTPS, SSL certificate setup and TCP port.

3. **Documentation Updates**:
   - Updated the `README.md` in the `bandersnatch_docker_compose` directory to include instructions for HTTPS support and volume configuration.
   - Added a test guide users to show how they can verify their HTTPS setup using either a web browser or the `curl` command.
   - Added steps pn how to generate a self-signed certificate for personal use.


pytest.ini
```diff
@@ -2,3 +2,4 @@
 log_cli_level = DEBUG
 log_level = DEBUG
 asyncio_mode=strict
+norecursedirs = .* src/bandersnatch_docker_compose
```
cooperlees, line 5: What does .* do here? Do we want that? Your goal is to just ignore your compose files right?

cooperlees (changes requested):
Cool - Thanks for the TLS support. I never added it as personally I'd use a dedicated box on top to do the TLS but i am sure this will be useful for people.

I only have a request to try understand why we're adding `.*` to norecursedirs for pytest - What was breaking? I would recommended explaining key changes to a projects CI configuration in future to make it easier for the reviewer. Thanks!

src/bandersnatch_docker_compose/README.md
```diff
@@ -33,24 +53,105 @@ docker compose logs -f bandersnatch
...
+> Note - Your browser may display a security warning if you are using a self-signed certificate.
+
+##### Using `curl` Command
+
+- Open a terminal or command prompt.
+
+- Run the following command:
+
+  ```bash
+  curl -k https://pypi-repo-domain:44300
```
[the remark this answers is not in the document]
benjisho: adding the -v is a good idea 😄
will help people get https logs right away in case of fail.

pytest.ini
benjisho: it will work, but the aim here is to avoid using `/` as it is a Linux directory format. So adding a `\` will not affect the result.

I will troubleshoot where exactly is the place with the issue in a separate pull request.
- Till then, i suggest we keep only the `norecursedirs = src\bandersnatch_docker_compose` to the `pytest.ini` so the test will ignore the compose directory.

CHANGES.md
```diff
@@ -2,6 +2,7 @@
 
 ## New Features
 
+- Added HTTPS support in Docker Compose + Enabled bind-mount volume for Nginx config + add documentation in README.md(PR #1653).
```
cooperlees, line 5: I'll fix it, but this should go into a unreleased version section

cooperlees (approved):

PR #1694, opened 2024-03-29
(Work-in-progress) Update config file reading and validation

### LINKED ISSUE #1674: Add SOCKS support to proxy configuration parameter
Doing a docs diff we found that the proxy bandersnatch.conf does not work correctly with SOCKS proxies. Let's tweak the code that checks for SOCKS URLs to check the proxy option in the config as well as the environment variables.

flyinghyrax:
## Goals

- Consolidate options from the `mirror` section into a type-annotated class
  - This expands on the existing `SetConfigValues` named tuple
  - Probably dataclass or attrs (I <3 attrs)
- Consolidate reading and validation of mirror options into the configuration module
  - Currently some options are read in `validate_config_values` and some in the `mirror` subcommand itself, we can move these into a single place
  - Similarly with validation, some validation is performed e.g. inside the Mirror/Master classes, we can validate configuration at the program boundary so internal classes can assume configuration is valid
- Define default values for all mirror options that are supposed to have one in a single location (#990)
  - Defaults should not have to be provided where an option is read, since this leads to defaults being set for some subcommands but not others 
- Change BandersnatchConfig into a regular class instead of a singleton
  - The config object is already being passed as a parameter from main to async_main to subcommands
  - " " " " " " " to the storage plugin initializers
  - The config object is not being passed to filter plugin initializers, but there is a clear place to do so
  - If the config object is injected as a parameter everywhere it is used, then a Singleton metaclass is unnecessary and all the unit tests become a bit simpler
- (If needed) support old and new config implementations to coexist and be used simultaneously so modules can be migrated over time
- Fix #1674 while we're in here

## In Progress

- [ ] A `MirrorConfig` attrs class
  - I've got most of an implementation for this, but intend to capture some design notes regarding the approach I ended up taking
- [ ] Update subcommands to to read values from validated class fields instead of directly from a `ConfigParser` instance
- [ ] ...

## Completed

- [x] Fix evaluation of custom option reference syntax in diff-file
- [x] Fix a diff file being written unconditionally 
- [x] Change storage plugin init to require an external ConfigParser instance 
- [x] Change filter plugin init to require " " " "

### DIFF
--- a/src/bandersnatch/tests/plugins/test_storage_plugin_s3.py
+++ b/src/bandersnatch/tests/plugins/test_storage_plugin_s3.py
@@ -144,7 +144,7 @@ def test_scandir(s3_mock: S3Path) -> None:
 
 
 def test_plugin_init(s3_mock: S3Path) -> None:
-    config_loader = mock_config(
+    config = mock_config(
         """
 [mirror]
 directory = /tmp/pypi
@@ -167,14 +167,14 @@ endpoint_url = http://localhost:9090
 signature_version = s3v4
 """
     )
-    backend = s3.S3Storage(config=config_loader.config)
+    backend = s3.S3Storage(config=config)
     backend.initialize_plugin()
 
     path = s3.S3Path("/tmp/pypi")
     resource, _ = path._accessor.configuration_map.get_configuration(path)
     assert resource.meta.client.meta.endpoint_url == "http://localhost:9090"
 
-    config_loader = mock_config(
+    config = mock_config(
         """
 [mirror]
 directory = /tmp/pypi
@@ -193,7 +193,7 @@ compare-method = hash
 endpoint_url = http://localhost:9090
 """
     )
-    backend = s3.S3Storage(config=config_loader.config)
+    backend = s3.S3Storage(config=config)
     backend.initialize_plugin()
 
     path = s3.S3Path("/tmp/pypi")
--- a/src/bandersnatch/tests/plugins/test_storage_plugin_s3.py
+++ b/src/bandersnatch/tests/plugins/test_storage_plugin_s3.py
@@ -1,3 +1,4 @@
+from configparser import ConfigParser
 from datetime import datetime
 
 from s3path import S3Path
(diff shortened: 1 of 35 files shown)


src/bandersnatch/tests/test_mirror.py
```diff
@@ -63,6 +66,16 @@ def iter_content(*args: Any, **kwargs: Any) -> Iterator[bytes]:
...
+MirrorFactory: TypeAlias = Callable[..., BandersnatchMirror]
+
+
+@pytest.fixture
+def mock_mirror_factory(tmp_path: Path, local_storage: Storage) -> MirrorFactory:
+    def _mock_mirror(**kwargs: Any) -> BandersnatchMirror:
+        return BandersnatchMirror(tmp_path, mock.Mock(), local_storage)
+
+    return _mock_mirror
+
```
flyinghyrax, line 78: Unused; remove

src/bandersnatch/tests/plugins/test_storage_plugin_s3.py
```diff
@@ -1,20 +1,31 @@
...
+_empty_cfg = ConfigParser()
+
+
+def s3_backend(bucket: S3Path) -> s3.S3Storage:
+    config = ConfigParser()
+    config.read_dict(
+        {"mirror": {"storage-backend": "s3", "directory": bucket.as_posix()}}
+    )
+    return s3.S3Storage(config=config)
+
```
flyinghyrax, line 18: Unused; remove

flyinghyrax:
...ran head-first into a good reason why `configparser.ExtendedInterpolation` was not already being used:

<detail>
<summary>Regular expressions also use the $ character</summary>

```
c:\Users\mrsei\source\bandersnatch\src\bandersnatch\tests\plugins\test_regex_name.py::TestRegexProjectFilter::test_plugin_check_match failed: self = <test_regex_name.TestRegexProjectFilter testMethod=test_plugin_check_match>

    def test_plugin_check_match(self) -> None:
        bc = mock_config(self.config_contents)
    
>       mirror = make_test_mirror(config=bc)

src\bandersnatch\tests\plugins\test_regex_name.py:98: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
src\bandersnatch\tests\plugins\util.py:27: in make_test_mirror
    local_config.read_dict(config)
...
    self._interpolate_some(parser, option, L, value, section, defaults, 1)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <configparser.ExtendedInterpolation object at 0x00000232263DFF90>
parser = <bandersnatch.config.core.BandersnatchConfig object at 0x00000232263DED10>
option = 'packages', accum = ['\n.+-evil'], rest = '$\n.+-neutral$'
section = 'filter_regex'
map = ChainMap({}, {'packages': '\n.+-evil$\n.+-neutral$'}, {}), depth = 1

    def _interpolate_some(self, parser, option, accum, rest, section, map,
                          depth):
        ...
        while rest:
            ...
            if c == "$":
                ...
            elif c == "{":
                ...
            else:
>               raise InterpolationSyntaxError(
                    option, section,
                    "'$' must be followed by '$' or '{', "
                    "found: %r" % (rest,))
E               configparser.InterpolationSyntaxError: '$' must be followed by '$' or '{', found: '$\n.+-neutral$'

C:\Program Files\Python311\Lib\configparser.py:517: InterpolationSyntaxError
```

</detail>

...these changes are getting large enough that I wonder if I need a rethink here.

PR #1697, opened 2024-04-02
feat: support arbitary boto3 s3 config options in s3 backend

joel-aws:


src/bandersnatch_storage_plugins/s3.py
```diff
@@ -123,6 +123,8 @@ class S3Storage(StoragePlugin):
     PATH_BACKEND = S3Path
     resource = None
     UPLOAD_TIME_METADATA_KEY = "uploaded-at"
+    CONFIG_PREFIX = "config_param_"
```
cooperlees, line 126:
```suggestion
    BOTO_CONFIG_PREFIX = "config_param_"
```
Maybe to make it clearer it's use

cooperlees (changes requested):
Hi and thanks for this. I am sure it will be helpful for S3 bandersnatch users.

Is there any easy way we can use our [minio unit/integration tests](https://github.com/pypa/bandersnatch/blob/main/src/bandersnatch/tests/plugins/test_storage_plugin_s3.py) to make use of any of these config params to see at least one work? I feel this will help with future API changes etc. etc.

src/bandersnatch_storage_plugins/s3.py
joel-aws: Agreed -- that's more clear. Will update.

joel-aws:
@cooperlees -- added a test that uses minio.

I also refactored `s3.py` to apply the boto3 configuration to all instantiation of `S3Path`. Though, when `.s3keep` objects are created, the configuration is *not* applied (couldn't figure that one out). Overall, I don't love my implementation, so I welcome suggestions.

cooperlees (approved):
I don't have suggestions no how to make this any cleaner. So I am happy to merge if you are.

> Though, when .s3keep objects are created, the configuration is not applied

Is there any open issue or is it worth opening an issue here to see if there is a bug or we're doing something wrong here?

If you fix the CHANGES entry I'll merge and if it benefits you cut a new release with the feature.

Thanks for the contribution! (I am no s3 guru tho)

cooperlees:
@joel-aws Happy for this to be merged?

joel-aws:
@cooperlees LGTM!

PR #1715, opened 2024-04-26
Fix custom config reference syntax for the `diff-file` option

flyinghyrax:
This replaces the implementation for custom `configparser` option references in the `diff-file` option, e.g. `diff-file = {{ section_option }}/myfilename.txt`.

Detection and handling for the custom syntax are moved into a submodule. The previous implementation did not preserve text before and after the reference part of the config value. This uses regular expressions to preserve any text before and after the reference.

Includes demonstration/regression tests added to the `bandersnatch/tests/configuration.py`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,10 @@
 
 - Updated documentation for `[mirror]` configuration options `PR #1669`
 
+## Big Fixes
+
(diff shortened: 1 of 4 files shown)


cooperlees (approved):
Many thanks!

Very nice all round. Docs, tests and good code comments with a great sized easy to review PR.

PR #1716, opened 2024-04-28
Fix diff-file being created when not configured

flyinghyrax:
This fixes path handling in the `mirror` subcommand that caused a diff file to be written to the current directory if `diff-file` wasn't set in the configuration file. This is the issue where `Path("")` becomes `Path(".")` and boolean-checks as truthy.

I encountered what might be an unexpected behavior when `diff-file` is set to an existing directory and `diff-append-epoch` is enabled. I've added a unit test to capture the current behavior, but it might be worth changing depending on how you feel about the compatibility break.

Split into 3 commits:

1. Removes a `diff_file` argument and attribute from `BandersnatchMirror`. The initializer accepted both `diff_file` and `diff_full_path`, and `diff_file` doesn't appear to be used anywhere. I went ahead and removed it while I was around then extracted this change into its own commit so it can be accepted/removed on its own.
2. Adds 2 tests for the mirror subcommand: 
    - check whether a diff file is being created when not configured
    - capture the interaction between the `diff-file` and `diff-append-epoch` options when `diff-file` is set to an existing directory

   If you check out this commit, the first test is expected to fail.
3. Changing the path handling in the mirror subcommand for `diff-file`.

### DIFF
--- a/src/bandersnatch/mirror.py
+++ b/src/bandersnatch/mirror.py
@@ -6,11 +6,11 @@ import logging
 import os
 import sys
 import time
-from collections.abc import Awaitable
+from collections.abc import Awaitable, Callable
 from json import dump
 from pathlib import Path, WindowsPath
 from threading import RLock
-from typing import Any
+from typing import Any, TypeVar
 from urllib.parse import unquote, urlparse
 
 from filelock import Timeout
@@ -22,7 +22,7 @@ from .filter import LoadedFilters
 from .master import Master
 from .package import Package
 from .simple import SimpleAPI, SimpleFormat, SimpleFormats
-from .storage import storage_backend_plugins
+from .storage import Storage, storage_backend_plugins
 
 LOG_PLUGINS = True
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


src/bandersnatch/tests/test_mirror.py
```diff
@@ -1297,5 +1300,170 @@ def test_write_simple_pages(mirror: BandersnatchMirror) -> None:
...
+    await mirror_cmd(config=config)
+
+    # Expectation: current behavior appends timestamp before checking if the directory exists,
+    # so the generated path will be `{diff-file}-{timestamp}`.
+    expected_diff_file = diff_files_dir.with_name(
+        f"{diff_files_dir.name}-{int(fixed_time)}"
+    )
+    assert expected_diff_file.is_file()
+    lines = expected_diff_file.read_text().splitlines()
+    assert len(lines) > 0
```
cooperlees, line 1465:
```suggestion
    assert len(lines) == 3
```
This should always be the same number of lines right? Or did I count wrong / miss why > 0 is better here?

cooperlees (changes requested):
LGTM - Just a couple NITs and questions ... If I misunderstood just explain so and we can look at accept and merge.

src/bandersnatch/mirror.py
```diff
@@ -913,6 +911,42 @@ async def download_file(
...
+
+
+async def _setup_diff_file(
+    storage_plugin: Storage, configured_path: str, append_epoch: bool
+) -> Path:
+    # save some typing. maybe a method to add to Storage?
+    async def storage_plugin_exec(fn: Callable[..., _T], *args: Any) -> _T:
+        return await storage_plugin.loop.run_in_executor(
+            storage_plugin.executor, fn, *args
+        )
```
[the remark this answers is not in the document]
flyinghyrax: Yeah, I thought this function might work as a method on the base class? I didn’t look to see how common using the storage plugin this way is in the rest of the code. I supposed that most of the time one would call a method on the storage backend directly. But if it could be useful elsewhere the storage plugins could have this as a method, just wrapping `run_in_executor` for convenience.

src/bandersnatch/tests/test_mirror.py
flyinghyrax: You didn’t miss anything, I think this was just laziness 😆 And after already mocking `synchronize` too.

I’ll double-check this and update it.

cooperlees: Good old Windows, I wonder if anything goes wrong on Linux and if you fix doesn't impact Linux/MacOS (POSIX filesystems) then I am cool for the fix to come in this PR.

flyinghyrax: I can do a before/after check in a docker container this weekend to make certain, since if it did affect Linux I'm not sure there's an existing unit test that would catch the regression in CI.

src/bandersnatch/mirror.py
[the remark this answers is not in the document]
flyinghyrax: Finally did a quick Ctrl+F and I'm sold on adding an `async def run_in_executor(...)` helper method to the `Storage` plugin base class, but I do think it would be better in a later PR.

cooperlees (approved):
Looks good to me - Lets fix other nits in follow up diffs!

PR #1753, opened 2024-06-18
Fix the command execution by runner.py

### LINKED ISSUE #1754: Runner.py cannot execute bandersnatch
When no --force-check option is specified when using the runner.py the bandersnatch command execution fails with: `bandersnatch: error: unrecognized arguments: `

vajdera:
When no `--force-check` option is specified when using the `runner.py` the argument list for running bandersnatch command consists of `''` at the end which breaks the bandersnatch command execution: `bandersnatch: error: unrecognized arguments: `. This commit fixes this by only appending the "--force-check" argument if specified in runner.py call.

Closes #1754


cooperlees (approved):
Hi. Don't know why it was done the other way, This is cleaner, so thanks.

PR #1760, opened 2024-06-27
docs: various tweaks

miketheman:
A few fixes, clarrifications

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,6 +45,7 @@ coverage.xml
 *,cover
 .hypothesis/
 test.conf
+local.conf
 
 # Translations
 *.mo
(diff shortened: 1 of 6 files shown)


miketheman:
As a docs-only, it's unclear why coverage is sad.

> Report is 117 commits behind head on main.

Is there something going on with how coverage on `main` is being uploaded (or not)? It doesn't look like the job has run on `main` - see https://github.com/pypa/bandersnatch/actions/runs/9648624809

ewdurbin:
https://github.com/pypa/bandersnatch/actions/runs/9648624821/job/26610268552#step:8:40 seems to be the culprit.

cooperlees (approved):
Many thanks for polishing it up.

I'll try dig into the coverage thing. Super confused there.

PR #1766, opened 2024-07-02
test: allow skipping S3-related tests

miketheman:
Show the user how to skip S3 tests.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,7 @@
 ## New Features
 
 - Add arbitrary configuration option for S3 Storage Backend Boto3 calls (PR #1697)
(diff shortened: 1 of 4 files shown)


miketheman:
I also thought a little bit about https://pypi.org/project/pytest-minio-mock/ - that might be a more interesting long-term idea, so no real minio server has to be running for tests.

cooperlees (approved):
Nice. Thanks for this. Will make easier for people not touching S3.

I would totally move a mocked library to make these unit tests again ... 
- But I worry about over mocking hiding issues these integration tests have found - e.g. auth setting bugs etc.

PR #1829, opened 2024-12-01
Upgrade test dependencies pytest and pytest-asyncio 

flyinghyrax:
## Dependency Upgrades

This performs the same version changes as #1656 and #1797 in a single merge request, since the associated issues and fixes appear to be related:

- `pytest` => 8.3.4
- `pytest-asyncio` => 0.24.0

This is a major version increment for pytest and a [0ver](https://0ver.org/) minor version increment for pytest-asyncio.

## Unit Test Changes

Locally, I used `pytest-randomly` to shuffle the test execution order. Varying the test order uncovered a number of places where shared state was causing order-dependent failures.

1. Tests that indirectly invoke a call to `get_running_loop`/`get_event_loop` require an event loop to be available even if the test itself and the code under test is not asynchronous. Depending on execution order an event loop was not always active when needed. I added `asyncio` marks to test modules wherever tests were failing because of this (or for 3.12, displaying a `DeprecationWarning` when there was no current event loop).

2. Storage plugin instances store the current event loop as an attribute, and we cache initialized plugins globally. This led to some situations where
    1. An event loop created for one test was accessed in another, and had been closed closed and replaced with a new event loop
    2. A storage plugin for a non-filesystem backend was re-used by a later test where the filesystem backend was expected.
  
   I created fixtures to clear cached storage plugins between test modules. Most problems seemed solved by clearing between modules and I made that an `autouse` fixture. For one function in `test_delete` the scope needed to be narrower and there's a comment inline explaining that case.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -2,6 +2,7 @@
 log_cli_level = DEBUG
 log_level = DEBUG
 asyncio_mode=strict
+asyncio_default_fixture_loop_scope = function
 
 testpaths = src
 python_files = test.py test_*.py
--- a/requirements_test.txt
+++ b/requirements_test.txt
@@ -1,13 +1,13 @@
-async-timeout>=4.0.0a3
+async-timeout==5.0.1
 black==24.4.2
 coverage==7.6.4
 flake8==7.1.1
 flake8-bugbear==24.8.19
 freezegun==1.5.1
 mypy==1.11.2
-pre-commit==4.0.0
-pytest==7.4.4
-pytest-asyncio==0.23.8
+pre_commit==4.0.0
+pytest==8.3.4
+pytest-asyncio==0.24.0
 pytest-timeout==2.3.1
 setuptools==72.2.0
 tox==4.18.0
--- a/src/bandersnatch/tests/conftest.py
+++ b/src/bandersnatch/tests/conftest.py
@@ -1,6 +1,7 @@
 # flake8: noqa
 import os
 import unittest.mock as mock
+from asyncio import AbstractEventLoop
 from collections.abc import Iterator
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Dict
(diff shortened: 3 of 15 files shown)


flyinghyrax:
The macOS and Linux test failures appear to show `test_delete.py` using the Swift storage plugin, and the swift http connection failing. That has the smell of the connection pool being created in a different event loop than the one used in `delete_path`. When I've got the time I'll look at getting a Swift container or something set up so I can debug it properly.

src/bandersnatch/delete.py
```diff
@@ -22,20 +22,19 @@
 
 async def delete_path(blob_path: Path, dry_run: bool = False) -> int:
     storage_backend = next(iter(storage_backend_plugins()))
+    loop = asyncio.get_event_loop()
     if dry_run:
         logger.info(f" rm {blob_path}")
         return 0
-    blob_exists = await storage_backend.loop.run_in_executor(
+    blob_exists = await loop.run_in_executor(
```
cooperlees, line 29: I bet the backend loop is being patch / mocked or something to make the tests work.

So I bet we don't want to do this change and workout why the backend's loop is failing here ...

cooperlees (changes requested):
I think we want to use the storage backend loop still ... If I get some time I'll try revert that and see what error we get since I think your key check of 
```
asyncio_default_fixture_loop_scope = function
```
was the big fix we needed as we need lots of event loops with how our tests are written ...

cooperlees:
I can only repro locally on my Mac running latest minio with py3.11 ... on py3.12 it works ... which is interesting ....

flyinghyrax:
I'm interested in adding `pytest-randomly` to test deps and changing CI to use `-p no:randomly` (so it wouldn't be active in CI) but I will make a separate PR. 🙂

cooperlees (approved):
Awesome work and many many thanks @flyinghyrax ... This is a huge cleanup with out test suite and will help keeping it cleaner in the future.

Also, love the comment explaining why stuff was done. Can't click merge fast enough!

I will try get a release out over the holiday break finally for 3.12 and start work on making bandersnatch work for 3.13,.

flyinghyrax:
You're welcome! I'm very happy that I was able to get things working with relatively small changes. 🙂

PR #1834, opened 2024-12-15
Accommodations for running tests in containers that don't match the host OS

flyinghyrax:
I have been working on running the tests within a container to better match the CI environment and more easily test with multiple Python versions. I've kept my Docker/Docker Compose configuration in a sibling folder but needed to make a handful of changes to run the tests successfully.

1. The tests didn't have a way to configure the endpoint URL used for S3 plugin tests, so I added a check for an environment variable to allow overriding `localhost:9000` with something else.
2. There was a newline handling issue with a filesystem storage plugin test that only occurs when `sys.platform` doesn't match the platform the project source files were checked out of source control on. My host is Windows, so a sample file had `\r\n` endings, but when running the test in a Linux container the test expects `\n`.
3. I had problems getting pytest to discover test files, but adding some explicit file discovery settings fixed it.

Each of the above is in its own commit with some additional description in the commit log messages.

Since these changes only modify the development environment I'm not sure if they need a corresponding changelog entry.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -2,7 +2,11 @@
 log_cli_level = DEBUG
 log_level = DEBUG
 asyncio_mode=strict
+
+testpaths = src
+python_files = test.py test_*.py
 norecursedirs = src/bandersnatch_docker_compose
+pythonpath = .
 
 markers =
     s3: mark tests that require an S3/MinIO bucket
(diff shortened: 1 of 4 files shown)


cooperlees (approved):
All seems sane. Thanks for making the simple PR and not putting it in with lots of changes to fix our async problems! Really appreciate that.

PR #1861, opened 2025-01-25
Add SOCKS support to proxy configuration parameter

### LINKED ISSUE #1674: Add SOCKS support to proxy configuration parameter

flyinghyrax:
This unifies proxy configuration handling for HTTP and SOCKS, so that:

- HTTP proxy URLs can be read from environment variables as well as the `mirror.proxy` option
- SOCKS proxy URLs can be read from the `mirror.proxy` option as well as environment variables

Fixes #1674

### DIFF
--- a/src/bandersnatch/config/proxy.py
+++ b/src/bandersnatch/config/proxy.py
@@ -46,7 +46,7 @@ def proxy_address_from_env() -> str | None:
     for proto in _supported_protocols:
         if proto in proxies_in_env:
             address = proxies_in_env[proto]
-            logger.debug("found %s proxy address in environment: %s", proto, address)
+            logger.debug("Found %s proxy address in environment: %s", proto, address)
             return address
     return None
 
@@ -74,11 +74,11 @@ def get_aiohttp_proxy_kwargs(proxy_url: str) -> Mapping[str, Any]:
     """
     lowered = proxy_url.lower()
     if lowered.startswith("socks"):
-        logger.debug("using SOCKS ProxyConnector for %s", proxy_url)
+        logger.debug("Using SOCKS ProxyConnector for %s", proxy_url)
         return {"connector": ProxyConnector.from_url(proxy_url)}
 
     if lowered.startswith("http"):
-        logger.debug("using HTTP proxy address %s", proxy_url)
+        logger.debug("Using HTTP proxy address %s", proxy_url)
         return {"proxy": proxy_url, "trust_env": True}
 
     return {}
--- a/src/bandersnatch/config/proxy.py
+++ b/src/bandersnatch/config/proxy.py
@@ -1,0 +1,84 @@
+"""
+Implements 2 aspects of network proxy support:
+
+1. Detecting proxy configuration in the runtime environment
+2. Configuring aiohttp for different proxy protocol families
+"""
+
+import logging
+import urllib.request
+from collections.abc import Mapping
+from typing import Any
+
+from aiohttp_socks import ProxyConnector
+
+logger = logging.getLogger(__name__)
+
+# The protocols we accept from 'getproxies()' in the an arbitrary but reasonable seeming precedence order.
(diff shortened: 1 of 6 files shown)


src/bandersnatch/config/proxy.py
```diff
@@ -0,0 +1,84 @@
...
+    This uses the protocol family of the URL to select one or the other and return the
+    corresponding keyword arguments in a dictionary.
+    """
+    lowered = proxy_url.lower()
+    if lowered.startswith("socks"):
+        logger.debug("using SOCKS ProxyConnector for %s", proxy_url)
+        return {"connector": ProxyConnector.from_url(proxy_url)}
+
+    if lowered.startswith("http"):
+        logger.debug("using HTTP proxy address %s", proxy_url)
```
cooperlees, line 81:
```suggestion
        logger.info("using SOCKS ProxyConnector for %s", proxy_url)
        return {"connector": ProxyConnector.from_url(proxy_url)}

    if lowered.startswith("http"):
        logger.info("using HTTP proxy address %s", proxy_url)
```
Wonder if we want to always tell people we're using a proxy?

src/bandersnatch/tests/test_proxy_config.py
```diff
@@ -0,0 +1,155 @@
...
+            "https://192.0.2.112",
+        ),
+        # prefer socks to http and socks5 to socks4
+        (
+            {
+                "HTTPS_PROXY": "https://192.0.2.112",
+                "SOCKS4_PROXY": "socks4://192.0.2.113:1080",
+                "SOCKS5_PROXY": "socks5://192.0.2.114:1080",
+            },
+            "socks5://192.0.2.114:1080",
```
cooperlees, line 36: Should we put in an IPv6 URL for good measure?

cooperlees (changes requested):
All looks great. Many thanks!

Just wondering if we should always log when we're applying proxy config and maybe test IPv6 URLs (they should) work ...

If you don't have time to test IPv6 and don't agree we should log cause somewhere else already does, just say so, that's fine :)

flyinghyrax:
> All looks great. Many thanks!
> 
> Just wondering if we should always log when we're applying proxy config and maybe test IPv6 URLs (they should) work ...
> 
> If you don't have time to test IPv6 and don't agree we should log cause somewhere else already does, just say so, that's fine :)

I think these are both excellent points! Ideally both should be straightforward. Thanks for the fast review!

src/bandersnatch/tests/test_proxy_config.py
flyinghyrax: I think that is a good idea - I was tripped up several times by `ProxyConnector.from_url` rejecting things I expected to be valid, so it's definitely worth verifying.

None of these new tests cover all the way through to when the `aiohttp.ClientSession` is constructed - do you think that is worth covering?

src/bandersnatch/config/proxy.py
flyinghyrax: Very good point, if someone configured a proxy (...or _didn't_ but one is used because of an environment variable), it would be important to know.

I left the messages in `bandersnatch.config.proxy` at 'debug', and added an 'info' level message to the Master initializer, with the idea that it's a little less verbose by default but can still optionally be traced in more detail.

src/bandersnatch/tests/test_proxy_config.py
cooperlees: No, I think there we should rely on aiohttp tests. But I just want to know we pass IPv6 correctly if the libraries support it (pretty sure they do).

cooperlees (approved):
All LGTM - Many thanks for the improvement and cleanup and extra testing.

Issue #1897, opened 2025-03-05
Replace deprecated XML-RPC Pypi APIs


pombredanne:
See https://mail.python.org/archives/list/pypi-announce@python.org/thread/5VOX33ARFQUYKIMKM5NS7PM7Z6ZNCSJY/ and:
- https://github.com/pypi/warehouse/issues/16642

> 
> The following PyPI XMLRPC methods are being permanently deprecated:
> 
> list_packages
> package_releases
> release_urls
> release_data

Technically https://warehouse.pypa.io/api-reference/xml-rpc.html#mirroring-support did not deprecated the `list_packages_with_serial` RPC, but in practice it times out and should likely be replaced with the new simple JSON API that provides a list of packages with serials 
See also https://docs.pypi.org/api/index-api/

@ewdurbin I assume that `list_packages_with_serial` is effectively deprecated even if not documented yet, correct?

ewdurbin:
No, that endpoint is still supported and the documentation is up to date.

The intention is to support the three methods `changelog_last_serial`, `changelog_since_serial`, and `list_packages_with_serial` until a better solution for mirroring API is developed, deployed, and adopted.

In this instance the JSON simple API is a great candidate for adoption to replace `list_packages_with_serial`, except for the caching aspect. XMLRPC `list_packages_with_serial` will give you an honest un-cached answer every time (which is why it takes 10-15 seconds). The JSON simple API endpoint will remain cached for 24 hours, which is why it is often nearly immediate.

Our metrics for that endpoint show that has been generally stable, with responses ranging from 20-30 seconds from the backends, and generally well below the 60 second internal timeout. Is there a timeout being set by bandersnatch's client libraries?

ewdurbin:
Looking into this, I did find a bit of a glaring performance issue in the `list_packages_with_serial` method (and others!).

Once deployed, that endpoint should get a bit faster. Turns out internally we were executing the DB query and dictionary build for response... twice 🙃

cooperlees:
O how I'd love to remove all XMLRPC from bandersnatch. I've tied. Long long ago. PR's welcome using other "approved" and preferable PEP'd APIs to get the same data if its:
- At last as efficient (client side - but up to add a little complexity/compute client side)
- Decently tested

[...]

PR #1898, opened 2025-03-05
Drop XMLRPC API to use PyPI Index API

pombredanne:
The XMLRPC is now obsolete and deprecated and even the parts that may
not be deprecated no longer work. Instead, let's use the simple Index
API with JSON to collect all packages and their changes.

Reference: https://github.com/pypa/bandersnatch/issues/1897
Signed-off-by: Philippe Ombredanne <pombredanne@nexb.com>


cooperlees (changes requested):
Many thanks! This is great and much cleaner / simpler to implement that I thought.
- My only worry is to see how long PyPI caches these pages ... cc: @ewdurbin - Where do we find this setting?
  - We should document that as XMLRPC had 0 caching ...
  - But maybe this is low as we've not had many problems fetching packages supplied via the XMLRPC to mirror/sync ...

I'm also sad I forgot this was a stretch goal that we put into a PEP I co-authored. This was all clever additions by @dstufft  <3.

Can we just fix typing, add info on the Simple API caching (if there is any) and that test name unless 1__1 means something I don't understand, but we don't have it anywhere else I think ...

cooperlees:
The speed increase is amazing too based on the integration tests - **>10s saved** :
- XMLRPC - https://github.com/pypa/bandersnatch/actions/runs/13679190278/job/38246889516
  - 27s
- SimpleJSON - https://github.com/pypa/bandersnatch/actions/runs/13686864483/job/38272166947?pr=1898
  - 10s

requirements_docs.txt
```diff
@@ -5,7 +5,6 @@ packaging==24.2
 requests==2.32.3
 sphinx==8.2.1
 MyST-Parser==4.0.1
-xmlrpc2==0.3.1
```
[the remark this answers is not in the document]
pombredanne: Actually, I just brutally looked for "rpc" and this looked unused. Does uv check for imports?

ewdurbin:
> * My only worry is to see how long PyPI caches these pages ... cc: @ewdurbin - Where do we find this setting?

Caching for endpoints is configured in their  view configurations, as "orgin_cache": For the `/simple/` route that is here: https://github.com/pypi/warehouse/blob/ef9978cdf279bc8cf953eec7092cc2846de6b982/warehouse/api/simple.py#L69-L73

That endpoint is one of the only ones we don't purge in response to changes, due to its poor performance on cold-cache. Most relevant endpoints are purged in more-or-less real-time on upload, such as `/simple/bandersnatch/` and `/pypi/bandersnatch/json`.

pombredanne:
@cooperlees re:
> So what do we want to do here?
> I think we should keep xmlrpc for now, leave it default, and allow a config param to use the simple API.
> Then we can look at getting the caching down and swap to default when we feel PyPI is ready 

I am OK with this... @ewdurbin any input on your side?

cooperlees:
Should I finish the PR and ship it with xmlrpc support on as default? I love the idea here.

Last ping for @ewdurbin - I know he gets a lot of pings :)

ewdurbin:
re-loading context...

ewdurbin:
I'm motivated to see this land. Being able to put XMLRPC on a timeline for complete deprecation is a dream!

If 1 hour is an acceptable TTL for the JSON variant of the /simple/ endpoint PyPI can support that.

cooperlees:
Just for all here understanding, how fresh are the xmlrpc endpoints for Bandersnatch today? Pretty much always to origin/main service and at most seconds out of date right?

So people mirroring can be at most 1 hour up to date with this change to simple API for calculating new packages to download?
- We will just have to document that for people wanting to cut over until we make it default here ...

ewdurbin:
> Just for all here understanding, how fresh are the xmlrpc endpoints for Bandersnatch today? Pretty much always to origin/main service and at most seconds out of date right?

There is zero caching. They are "at time of database transaction". Effectively immediate.

> So people mirroring can be at most 1 hour up to date with this change to simple API for calculating new packages to download?

When using JSON /simple/, Once referenced warehouse PR is merged, yes.

pombredanne:
@ewdurbin 
> If 1 hour is an acceptable TTL for the JSON variant of the /simple/ endpoint PyPI can support that.

quite acceptable to me!

@cooperlees 

what's my next step then?

ewdurbin:
FWIW I don't think we're going to go much lower than what was merged earlier (30 minute TTL). So if that's not acceptable long term, I think we should wait until there's a more acceptable "real-time" API.

cooperlees:
> what's my next step then?

Restoring default support for xmlrpc controlled via configuration. I feel we can release noting we are planning to deprecate xmlrpc support, advertise the 30min limitation of simple API, but state how it's faster and cheaper for PyPI to run etc. but you lose sub 30min updating ability ...

I am sure >=30 mins is fine for 99% of our users ... but I have no way of knowing ...

blg-666 (approved):

src/bandersnatch/master.py
```diff
@@ -141,58 +136,47 @@ async def url_fetch(
...
-        if not all_packages_with_serial:
-            raise XmlRpcError("Unable to get full list of packages")
-        return all_packages_with_serial
+    async def fetch_simple_index(self) -> Any:
+        """Return a mapping of all project data from the PyPI Index API"""
+        logger.debug(f"Fetching simple JSON index from {self.simple_url}")
+        async with self.session.get(
+            self.simple_url, headers=self._custom_headers
+        ) as response:
+            simple_index = await response.json()
```
pelson, line 157: There is no guarantee that the return is actually JSON according to the spec https://peps.python.org/pep-0691/#content-types.

cooperlees: Good call, we did that more for migration and writing defensive code. So we should wrap this and error saying didn't get JSON when we asked for it so the error is clear ...

cooperlees:
@pombredanne - I've put up #2075 so we can ship this support non default and discuss moving it to default after that ...

Issue #2090, opened 2025-11-27
Fix docs theme


cooperlees:
https://bandersnatch.readthedocs.io/en/7.0.0/

It's all misformatted and ugly. Would love any help fixing it.

PR #2092, opened 2025-11-27
Use Furo theme for docs

hugovk:
Fixes https://github.com/pypa/bandersnatch/issues/2090.

I think the problem is with the PyPA Sphinx Theme theme, and the repo https://github.com/pypa/pypa-docs-theme was archived last year.

Instead, we could use something a bit more modern and, more to the point, maintained such as Furo: https://github.com/pradyunsg/furo


cooperlees:
Much more readable! Thanks again.

Issue #1451, opened 2023-05-03
Generate SimpleDigests object from what metadata offers


cooperlees:
Today in [simple.py](https://github.com/pypa/bandersnatch/blob/main/src/bandersnatch/simple.py) we hard code a NamedTuple of possible digests and a StrEnum equivalent. I'd like to see if we could refactor this to support any digest that PyPI itself offers in the metadata we use. Bandersnatch (unfortunately) uses the non PEP standard "JSON API".

An example metadata can be seen looking at our [bandersnatch](https://pypi.org/pypi/bandersnatch/json) project metadata. We can see here that it offers multiple hashes per release. We should allow users of bandersnatch to select any Digest offered there that is PEP691 compliant, i.e. hashlib supports.

This will help us just magically support the next new hash hotness when one of these or all become insecure.

PR #2262, opened 2026-06-04
Remove hardcoded SimpleDigest enum and validate digest against hashlib (issue #1451)

raghuvardhan07:
The `SimpleDigest` StrEnum hardcoded sha256 and md5 as the only valid
digest names. This replaces it with a check against `hashlib.algorithms_available`,
so any algorithm PyPI starts offering that hashlib supports will work without
a code change.

One non-obvious decision: `InvalidDigestFormat` now inherits `Exception` instead
of `ValueError`. The reason is that removing the enum means an invalid-but-hashlib-
valid digest (e.g. sha512) passes startup validation but fails at runtime when
the source index doesn't provide it. `BandersnatchMirror.on_error` has a pre-
existing `isinstance(exception, ValueError): pass` that would silently swallow this. `configuration.py`'s startup catch is updated to `except InvalidDigestFormat` accordingly.

`SimpleAPI.validate_digest_availability()` is called in `process_package` before
downloading files so a misconfigured digest fails immediately rather than after
pulling everything down.

Because we removed the hardcoded digests, validation will happen at process_package instead of initialization thus burning through some bandwidth if the digest is incorrectly configured

### DIFF
--- a/src/bandersnatch/simple.py
+++ b/src/bandersnatch/simple.py
@@ -1,7 +1,8 @@
+import hashlib
 import html
 import json
 import logging
-from enum import Enum, StrEnum, auto
+from enum import Enum, auto
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, NamedTuple
 from urllib.parse import urlparse
@@ -23,16 +24,6 @@ class SimpleFormat(Enum):
     JSON = auto()
 
 
-class SimpleDigests(NamedTuple):
-    sha256: str
-    md5: str
-
-
-class SimpleDigest(StrEnum):
-    SHA256 = "sha256"
-    MD5 = "md5"
-
-
 logger = logging.getLogger(__name__)
 
 
@@ -42,7 +33,7 @@ class InvalidSimpleFormat(KeyError):
     pass
 
 
-class InvalidDigestFormat(ValueError):
+class InvalidDigestFormat(Exception):
     """We don't have a valid digest choice from configuration"""
 
     pass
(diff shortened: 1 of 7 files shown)


src/bandersnatch/simple.py
```diff
@@ -42,7 +33,7 @@ class InvalidSimpleFormat(KeyError):
     pass
 
 
-class InvalidDigestFormat(ValueError):
```
cooperlees, line 45: What happens if this stays a subclass of ValueError? I'm missing that

cooperlees (commented):
O nice, yeah we do expose this and I don't think it's PEP691 but somewhere states PyPI can only use hashes that are supported by hashlib. So this is a good improvement. Thanks! 

Let's see what copilot says too ... post that I'll approve.

cooperlees:
P.s. fix black and copilot's suggestion unless you don't agree ...

raghuvardhan07:
> P.s. fix black and copilot's suggestion unless you don't agree ...

Ive fixed the lint, i understood the copilot suggestion and agree with it. some user might have configured "Sha256" and it would have worked for them as the old function does upper() on the string. Now im doing lower() and matching.

src/bandersnatch/simple.py
raghuvardhan07: The Bandersnatch class has an on_error method which intentionally skips ValueError and TypeError for some reason. (around line 370 in mirror.py). The reason this wasnt an issue before is that digest validation happened in validate_config_values() before the mirror class was initialized. If a digest like sha512 is passed, it will pass the validate_config_values() check as its part of hashlib, but it will only be caught in process_package whose exceptions are passed into on_error. So in order for the InvalidDigestName to be thrown correctly, it has to not be ValueError hence the change. I've made sure this change reflects in every path.

cooperlees: Interesting and nice find.

cooperlees (approved):
Seems amazon changes their docs - So CI failure not related to this PR. So merging and I/someone will have to fix those 403 errors in the docs build.

Issue #2264, opened 2026-06-06
Move all tests to be pytest


cooperlees:
Since we have pytest, we should use it everywhere. So let's convert all the unittest module tests to be pytest.

PR #2265, opened 2026-06-07
Refactor tests from unittest to pytest (issue #2264)

raghuvardhan07:
Refactored top level src tests and the `test_docker_runner.py` to use pytest completely and follow its testing principles. Added `pytest-mock` dependency to use its patcher method which is much more convenient and safe than doing patch.start() and patch.stop()

Moved s3 related test imports inside the test itself and used importorskip as the whole s3 testing situation has a dedicated requirements folder, its own test file etc. Importing them at the top of the src conftest means if those imports are missing, the whole suite crashes.

Added fixture for tmp directory handling in `test_filter.py`

Also changed CaptureFixture to LogCaptureFixture in test_package.py as those functions use the logger methods instead of std.

(I have 2 other chained PRs where ive attempted to refactor the plugins test files as well)

Love to here your thoughts on this refactor attempt and see if any changes are required.

Fixes ##2264

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,7 @@
 
 - Declare workflow-level `permissions: contents: read` on the 8 remaining workflows (CVE-2025-30066 hardening) `PR #2241`
 - Skip base class tests in test runs `PR #2263`
+- Transitioned from unittest.Testcase to pytest fixtures `PR #2265`
 
 ## Documentation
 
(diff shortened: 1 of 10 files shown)


cooperlees (approved):
All looks good to me. I'm no pytest expert but these all look to be working to em and codecov. Thanks!

PR #2273, opened 2026-06-09
Refactor plugin tests

raghuvardhan07:
Added common tmp directory handling fixture which also sets a clean loaded_storage_plugins object to bandersnatch in tests/plugins/conftest.py. Moved from unittest and class based testing to pytest style tests and fixtures.


cooperlees (approved):
Thanks! Forgot these were unittest too. Nice.

PR #2274, opened 2026-06-10
Refactor test_storage_plugin.py

raghuvardhan07:
Moved from class based hierarchial tests to pytest params based tests. added a backend() fixtures that effectively makes the whole set of tests run for all the params we mention for that fixture (currently it only tests 'filesystem' backend like before). Because class based tests are removed, instead of using self, i made a dataclass called storage_env that comes in place of that. consolidated multiple setup_[action] functions into one function. parameterized as much as possible to avoid internal for loops.


cooperlees (approved):
Awesome. Thanks again!

Issue #2052, opened 2025-10-03
Feature Request: Automatic deletion of previously mirrored files excluded by filtering configuration


naveen1583:
## Summary

Currently, when filtering configuration is changed to exclude certain files (e.g., using `exclude_platform` to filter out old Python versions), bandersnatch does not automatically delete previously mirrored files that are now excluded by the active filters. This results in wasted storage space and creates a mismatch between user expectations and actual mirror contents.

## Problem

When users update their filtering configuration to be more restrictive (for example, excluding packages for Python 2.7 or old platform-specific wheels), they expect that subsequent mirror synchronizations would remove files that no longer match the active filter criteria. However, bandersnatch currently only prevents new downloads of filtered content but leaves previously downloaded files in place.

This leads to:
1. **Wasted storage space** - Old files that are no longer needed continue to consume disk space
2. **Inconsistent mirror state** - The mirror contains files that would not be downloaded if starting fresh with the current configuration
3. **Manual cleanup burden** - Users must manually identify and delete files, which is error-prone and time-consuming

## Proposed Solution

Add functionality to automatically delete previously mirrored files that are now excluded by the active filtering configuration. This could be implemented as:

1. **Automatic deletion during sync** - An optional configuration flag (e.g., `cleanup_filtered_files = true`) that enables automatic removal of files excluded by current filters during the mirror synchronization process

2. **Separate cleanup command** - A dedicated command (e.g., `bandersnatch cleanup-filtered`) that scans the mirror and removes files that don't match the current filter configuration

## Use Cases

- Filtering old Python versions (e.g., Python 2.7 packages) to save storage after they're no longer needed
- Excluding specific platforms using `exclude_platform` filters
- Adjusting package allowlists/blocklists and wanting the mirror to reflect those changes
- Applying regex or metadata filters and needing historical files to be cleaned up

## Benefits

- Automatic storage optimization
- Mirror state consistent with current configuration
- Reduced manual maintenance overhead
- User expectations met when changing filter settings

## Related Issues

This is related to general package cleanup discussions in the project but specifically focuses on the interaction between filtering configuration and file retention.

Henko551:
[materials.zip](https://github.com/user-attachments/files/23507927/materials.zip)

PR #2277, opened 2026-06-10
Optimize hash method

raghuvardhan07:
Replace hashing implementation based on `Path.read_bytes()` with `hashlib.file_digest()` in the hash() utility.
This changes hashing from loading the entire file into memory to streaming directly from the file object.
(Would speed up the verify action)

I stumbled on this when looking at this issue: (issue #2052)
As bandersnatch now requires python 3.12 and here we can use this new function which is significantly faster

My benches:

os: windows

i generated 3 files (100mg, 1gb and 5gb) filled with random bytes
Then ran this bench script: 
[bench_hash.py](https://github.com/user-attachments/files/28801714/bench_hash.py)
I observed that the existing hash method took 50 seconds on first runs and averaged 22 seconds in subsequent runs (probably due to OS hot caches)
Whereas the file_digest didnt cross 6 seconds. feel free to run the inspect the bench script and run it. 

The reason this does well is the existing path.read_bytes() loads the entire file in mem. file_digest() streams and has a dedicated C implementation for it.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,7 @@
 ## New Features
 
 - Allow configuring any PEP691 compliant digest that the index offers `PR #2262`
+- Use hashlib.file_digest() to calculate file hashes when verifying `PR #2277`
 
 ## CI / test
 
--- a/src/bandersnatch/utils.py
+++ b/src/bandersnatch/utils.py
@@ -59,12 +59,11 @@ def convert_url_to_path(url: str) -> str:
 
 
 def hash(path: Path, function: str = "sha256") -> str:
-    h = getattr(hashlib, function)
-    content = path.read_bytes()
-    result = h(content).hexdigest()
-    if isinstance(result, str):
-        return result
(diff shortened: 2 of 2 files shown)


cooperlees:
O wow, this is an awesome find! Thanks for noticing it and putting this PR up.

cooperlees (approved):
This is great to reduce memory usage and perf!

raghuvardhan07:
I had a design question. In prod setups why would someone just use the `verify` command instead of `verify --delete`. verify is meant to be a hard sync right? Why not make delete the default and no delete the optional flag? @cooperlees

cooperlees:
> I had a design question. In prod setups why would someone just use the `verify` command instead of `verify --delete`. verify is meant to be a hard sync right? Why not make delete the default and no delete the optional flag? @cooperlees

Mainly for a "dry run" mode ... so you can inspect / see what could be deleted etc. and as a safety net so people only delete if they actually want to delete files ...

PR #2279, opened 2026-06-12
fix: reset --force-check serial via storage backend (closes #2278)

### LINKED ISSUE #2278: Force check logic has a potential bug
```
if args.force_check:
        storage_plugin = next(iter(storage_backend_plugins()))
        status_file = (
            storage_plugin.PATH_BACKEND(config.get("mirror", "directory")) / "status"
        )
        if status_file.exists():
            tmp_status_file = Path(gettempdir()) / "status"
            try:
                shutil.move(str(status_file), tmp_status_file)
                logger.debug(
                    "Force bandersnatch to check everything against the master PyPI"
                    + f" - status file moved to {tmp_status_file}"
                )
            except OSError as e:
                logger.error(
                    f"Could not move status file ({status_file} to "
                    + f" {tmp_status_file}): {e}"
                )
        else:
            logger.info(
                f"No status file to move ({status_file}) - Full sync will occur"
            )
```

The use of shutil.move() assumes that status file is a local path which it wont be for storage plugins like s3 and stuff. Am i right in interpreting this? 

Also why are we "moving" the status file to a tmp file instead of writing 0 to the existing one?

enestanrikuluu:
## What

`mirror --force-check` reset the PyPI serial by `shutil.move()`-ing the local `status` file to `/tmp`. That assumes the status file is on the local filesystem, so for non-local storage backends (e.g. S3) force-check fails / doesn't reset the serial as intended.

This resets the serial to `0` through the configured storage backend's `write_file()` instead, so a full sync is forced regardless of backend (`synced_serial == 0` already triggers a full sync in `mirror.py`).

Per @cooperlees on #2278: "Move or 0 will both work, I'd accept a PR moving to a write 0 using how ever we write it today for the respective backends too."

## Changes

- `main.py`: replace the `shutil.move()`-to-`/tmp` logic with `storage_plugin.write_file(status_file, "0")`; drop now-unused `shutil` / `gettempdir` imports.
- Update the `--force-check` help text (it no longer moves the file to /tmp).
- Add regression tests for the existing-status-file path (asserts it is reset to `0` in place, not moved) and the missing-status-file path.
- `CHANGES.md` entry under Bug Fixes.

## Testing

- New + existing `test_main.py` pass (8/8); 107 passing across `test_main` + storage-plugin tests.
- The new regression test fails on `main` (status file is moved away) and passes with this change.
- `black`, `isort`, `flake8`, `mypy`, `mdformat` all clean.

## AI assistance disclosure

This change was prepared with AI assistance (Claude); I reviewed and verified the diff and test results before submitting.

Closes #2278

### DIFF
--- a/src/bandersnatch/main.py
+++ b/src/bandersnatch/main.py
@@ -2,11 +2,9 @@ import argparse
 import asyncio
 import logging
 import logging.config
-import shutil
 import sys
 from configparser import ConfigParser
 from pathlib import Path
-from tempfile import gettempdir
 
 import bandersnatch.configuration
 import bandersnatch.delete
@@ -62,8 +60,8 @@ def _mirror_parser(subparsers: argparse._SubParsersAction) -> None:
         action="store_true",
         default=False,
         help=(
-            "Force bandersnatch to reset the PyPI serial (move serial file to /tmp) to "
-            + "perform a full sync"
+            "Force bandersnatch to reset the PyPI serial (set the status file to 0) "
+            + "to perform a full sync"
         ),
     )
     m.set_defaults(op="mirror")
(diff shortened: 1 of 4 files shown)


src/bandersnatch/main.py
```diff
@@ -171,22 +169,21 @@ async def async_main(args: argparse.Namespace, config: ConfigParser) -> int:
...
             try:
-                shutil.move(str(status_file), tmp_status_file)
+                # Reset the serial to 0 via the storage backend so a full sync
+                # occurs. Using the backend keeps this working for non-local
+                # backends (e.g. S3), unlike a local shutil.move(). See #2278.
+                storage_plugin.write_file(status_file, "0")
                 logger.debug(
                     "Force bandersnatch to check everything against the master PyPI"
-                    + f" - status file moved to {tmp_status_file}"
+                    + f" - status file {status_file} reset to serial 0"
```
cooperlees: Let's read and log the old serial number
```suggestion
                    + f" - status file {status_file} was reset to serial 0 (from {before_serial})"
```

cooperlees (changes requested):
Thanks for this - Let's just add logging of the old serial in case someone needs it for some reason since we're overwriting it ...

+ make changelog and pre-commit happy.

cooperlees:
O my bad, I'm sorry CLAUDE pushed bad markdown, I thought it was in the README/CHANGELOG ... I should have fixed that and had you rebase!

cooperlees (approved):

PR #2280, opened 2026-06-12
Fix broken symlinks with trailing newlines in targets

samwgoldman:
## Summary

Two tracked symlinks in `src/bandersnatch_docker_compose/` have a trailing newline byte baked into their targets, which makes them broken on disk even though the target files exist:

- `src/bandersnatch_docker_compose/Dockerfile` -> `../banderx/Dockerfile\n`
- `src/bandersnatch_docker_compose/nginx.conf` -> `../banderx/nginx.conf\n`

The intended targets (`src/banderx/Dockerfile`, `src/banderx/nginx.conf`) exist. The newline causes tools that follow symlinks while walking the tree to fail with `No such file or directory` pointing at these entries.

This PR recreates both symlinks with correct targets (no trailing newline). File mode stays `120000` (symlink).

## Test plan

Run from the repo root. `test -e` follows the symlink and only succeeds if it resolves to an existing file.

**Before** (symlink targets contain a trailing newline):

```
$ git cat-file -p HEAD~1:src/bandersnatch_docker_compose/Dockerfile | xxd | tail -1
00000010: 6572 6669 6c65 0a                        erfile.
#                            ^^ trailing 0a (newline)

$ for f in src/bandersnatch_docker_compose/Dockerfile \
           src/bandersnatch_docker_compose/nginx.conf; do
    test -e "$f" && echo "$f RESOLVES" || echo "$f BROKEN"
  done
src/bandersnatch_docker_compose/Dockerfile BROKEN
src/bandersnatch_docker_compose/nginx.conf BROKEN
```

**After** (no trailing newline):

```
$ git cat-file -p HEAD:src/bandersnatch_docker_compose/Dockerfile | xxd | tail -1
00000010: 6572 6669 6c65                           erfile
#                            no trailing 0a

$ for f in src/bandersnatch_docker_compose/Dockerfile \
           src/bandersnatch_docker_compose/nginx.conf; do
    test -e "$f" && echo "$f RESOLVES" || echo "$f BROKEN"
  done
src/bandersnatch_docker_compose/Dockerfile RESOLVES
src/bandersnatch_docker_compose/nginx.conf RESOLVES
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)


cooperlees (approved):
Thanks for fixing. Wonder how I / someone did this …

PR #2281, opened 2026-06-14
adds new project filter based on number of releases

### LINKED ISSUE #1771: Any way we can filter by the total number of versions?
Hi again, I'm trying to reduce the mirror size because now the deletion takes too long.
(approx 1 day with xargs + rsync)

I've found that there are 440k packages that only have 2 versions, which are quite obviously not used by anyone else but the authors themselves, and want to find a way to filter them.

Is there any way we can do this?

image

raghuvardhan07:
Adds a new metadata plugin to filter projects by number of releases. Closes #1771

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 - Remove legacy XML-RPC support paths; metadata sync now uses only the PEP 691 Simple API `PR #2276`
 - Allow configuring any PEP691 compliant digest that the index offers `PR #2262`
 - Use hashlib.file_digest() to calculate file hashes when verifying `PR #2277`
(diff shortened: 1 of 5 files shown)


cooperlees (approved):
Thanks! This all looks good to me - Testing makes sense. Will merge once I get all the deps landed and see tests stay passing.

cooperlees (changes requested):
Sorry, but copilot is smarter than me and spotted some nice things to polish up!

raghuvardhan07:
ahh. overlooked it. ive made the changes copilot gave + some other cases it missed (setting intialized to true in those errorneous cases) and added more tests

cooperlees (approved):

Issue #1124, opened 2022-05-05
Discuss: Support Core Metadata Mirroring


cooperlees:
PEP 658 specifies serving core package metadata from an index. Lets look at best ways to potentially mirror this is there is a sane way.

For more info: https://peps.python.org/pep-0658/

## Ideas

- Does bandersnatch have a config option to pull down the files to be statically served 
- Should this be left to the mirror service to pull the files out of the downloaded binaries and cached?

Other ideas?

pombredanne:
@cooperlees Hi :)

> Does bandersnatch have a config option to pull down the files to be statically served
> Should this be left to the mirror service to pull the files out of the downloaded binaries and cached?

I could see a config named `metadata` similar to the `release-files` and `json`  and essentially have the same behaviour as `json`. 

An alternative is to deprecate the `json` config and have a new `json-and-metadata` config that would pull both or none and could have this documentation:

> The mirror json-and-metadata setting is a boolean (true/false) setting that indicates that the json packaging metadata and the core metadata (pep-0658) should be mirrored in addition to the packages.

pombredanne:
FWIW, https://peps.python.org/pep-0714/ supersedes https://peps.python.org/pep-0658 and is deployed (and all populated now). For instance https://files.pythonhosted.org/packages/3f/02/6389ef0529af6da0b913374dedb9bbde8eabfe45767ceec38cc37801b0bd/boolean.py-4.0-py3-none-any.whl.metadata

xiaoyao9184:
So, it's still not supported yet?

cooperlees:
Mirroring PEPs 0658/0714 has not been done or attempted by anyone (to my knowledge). PR(s) welcome.

cooperlees:
The easiest way to support this I feel is to just add pulling each metadata file and saving for the mirror, like how we do the "JSON API" ... Then the file can be served with any headers needed etc. with nginx or server of choice ...

raghuvardhan07:
Downloading the metadata file for the dist means we have to check whether it exists and have to verify hash. We'd have to hit the /simple API for that for each package.

```
{
            "core-metadata": {
                "sha256": "8300940ef4a13f0aa2cfb9e3af3508bc5f70bc9d1675469511fe984f0c8710e2"
            },
            "data-dist-info-metadata": {
                "sha256": "8300940ef4a13f0aa2cfb9e3af3508bc5f70bc9d1675469511fe984f0c8710e2"
            },
[...]

PR #2288, opened 2026-06-19
Update get hash and others

raghuvardhan07:
- get_hash() in file_system implements the same hashlib.file_digest as in utils hash() method.
- LoadedFilters() was initialized twice, moved the init outside
- Made a clean classmethod for package class to create a package object from a pkg dict

### DIFF
--- a/src/bandersnatch/verify.py
+++ b/src/bandersnatch/verify.py
@@ -167,13 +167,15 @@ async def verify(
         with storage_backend.open_file(json_full_path, text=True) as jfp:
             metadata = json.load(jfp)
     except json.decoder.JSONDecodeError as jde:
-        logger.error(f"Failed to load {json_full_path}: {jde} - skipping ...")
+        logger.error(f"Failed to load {json_full_path} metadata: {jde} - skipping ...")
         return
 
     try:
         pkg = Package.from_metadata(metadata)
     except ValueError as e:
-        logger.error(f"Failed to load {json_full_path}: {e} - skipping ...")
+        logger.error(
+            f"Failed to load {json_full_path} into a Package: {e} - skipping ..."
+        )
         return
 
     # apply releases filter plugins like class Package
--- a/src/bandersnatch/verify.py
+++ b/src/bandersnatch/verify.py
@@ -165,18 +165,21 @@ async def verify(
 
     try:
         with storage_backend.open_file(json_full_path, text=True) as jfp:
-            pkg = json.load(jfp)
+            metadata = json.load(jfp)
     except json.decoder.JSONDecodeError as jde:
         logger.error(f"Failed to load {json_full_path}: {jde} - skipping ...")
         return
 
+    try:
+        pkg = Package.from_metadata(metadata)
+    except ValueError as e:
+        logger.error(f"Failed to load {json_full_path}: {e} - skipping ...")
+        return
+
     # apply releases filter plugins like class Package
-    pkg_c = Package(pkg["info"]["name"])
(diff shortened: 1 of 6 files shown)


src/bandersnatch/verify.py
```diff
@@ -165,18 +165,21 @@ async def verify(
...
     except json.decoder.JSONDecodeError as jde:
         logger.error(f"Failed to load {json_full_path}: {jde} - skipping ...")
         return
 
+    try:
+        pkg = Package.from_metadata(metadata)
+    except ValueError as e:
+        logger.error(f"Failed to load {json_full_path}: {e} - skipping ...")
+        return
+
```
cooperlees, line 180: Much cleaner - Thanks!

Let's make the log unique (even tho we'll get log line) - Anything you can do to help someone debugging is worth it imo!
```suggestion
            metadata = json.load(jfp)
except json.decoder.JSONDecodeError as jde:
logger.error(f"Failed to load {json_full_path} metadata: {jde} - skipping ...")
return

    try:
        pkg = Package.from_metadata(metadata)
    except ValueError as e:
        logger.error(f"Failed to load {json_full_path} into a Package: {e} - skipping ...")
        return

```

src/bandersnatch_storage_plugins/filesystem.py
```diff
@@ -268,19 +268,15 @@ def is_file(self, path: PATH_TYPES) -> bool:
...
             path = pathlib.Path(path)
         logger.debug(
             f"Opening {path.as_posix()} in binary mode for hash calculation..."
         )
         with open(path.absolute().as_posix(), "rb") as f:
-            for chunk in iter(lambda: f.read(128 * 1024), b""):
-                logger.debug(f"Read chunk: {chunk!s}")
-                h.update(chunk)
-        digest = h.hexdigest()
+            digest = hashlib.file_digest(f, function).hexdigest()
```
cooperlees, line 277: This should help memory usage right? Verify was always a big memory user so thanks!

cooperlees (changes requested):
Thanks for this cleanup. Just a few small things to clean up that I and copilot found.

Love all your work. Feel free to keep them coming <3

I might try to releae 8.0.0 today - I got codecov fixed today.

src/bandersnatch_storage_plugins/filesystem.py
```diff
@@ -268,19 +268,15 @@ def is_file(self, path: PATH_TYPES) -> bool:
...
         logger.debug(f"Calculated digest: {digest!s}")
-        return str(h.hexdigest())
+        return str(digest)
```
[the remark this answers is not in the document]
raghuvardhan07: The original code type casted it. I didnt have the string type cast until i saw the diff and thought why not keep the same behaviour

src/bandersnatch_storage_plugins/filesystem.py
raghuvardhan07: Yes, the same method is used in the hash() utility method

cooperlees (approved):
Awesome - so fast! Many thanks.

raghuvardhan07:
Is PEP 658/714 support in the priority list? ive been looking to ways to implement it and its not simple. becoz we use warehouse api, we dont have the simple api's .metadata hash. Is it worth the effort do u think? @cooperlees

cooperlees:
I'd like to support it, but no one has jumped up and down for it - Let's discuss how to implement it on https://github.com/pypa/bandersnatch/issues/1124

